PR #2, opened 2016-10-28
Check error descriptions during untyped tests

jdub:
- This test will pass on current master, which always uses buffered
  machine registers for syscalls

- It will fail if syscall_stub_gen.py is run without --buffer, as
  the error descriptions are lost on the way through [1]

- It will pass once more when seL4/seL4#45 lands :-)

[1] http://sel4.systems/pipermail/devel/2016-October/001079.html

### DIFF
--- a/apps/sel4test-tests/src/tests/inc_untyped.c
+++ b/apps/sel4test-tests/src/tests/inc_untyped.c
@@ -28,6 +28,17 @@ test_retype(env_t env)
     error = vka_alloc_untyped(&env->vka, seL4_TCBBits + 3, &untyped);
     test_assert(error == 0);
 
+    /* Try to insert 0. */
+    error = seL4_Untyped_Retype(untyped.cptr,
+                                seL4_TCBObject, 0,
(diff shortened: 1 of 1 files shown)


jdub:
Made tests 32/64 bit safe. 😳

AdrianDanis:
This looks good to me. Want to squash the second commit into the first and then I'll push it through?

jdub:
Oh, does your Bitbucket workflow mean you can't squash merge over here?

AdrianDanis:
I can do that, but I don't think github can then realize that this PR should be closed, since these exact commit hashes will not exist in the history

jdub:
Yeah, you can now choose how to merge the PR: https://github.com/blog/2141-squash-your-commits

AdrianDanis:
Ahh I misunderstood your question. Yeah so our workflow is to merge this in our bitbucket repos, and then push that back out to github. If we merge on github then bitbucket/github can get out of sync.

One day we might be able to have github be the single canonical version but there are various complicated and silly reasons why we cannot right now.

jdub:
Squashed!

PR #4, opened 2016-10-30
Test range max against defined number

jdub:


jdub:
Couldn't figure out how to combine a preprocessor constant and integer suffix, so switched to test_assert.

Smattr:
You might have some success with something like `JOIN(CONFIG_RETYPE_FAN_OUT_LIMIT, ul)` using the definition from utils/stringify.h.

AdrianDanis:
Can also just do an explicit cast like ` (unsigned long)CONFIG_RETYPE_FAN_OUT_LIMIT`

jdub:
Thanks -- the explicit cast seemed neater. Tested on sabre/qemu and x64/qemu.

PR #5, opened 2016-12-06
Bring Bamboo / simulation configurations closer

jdub:
- Disable X2APIC for simulation builds, same as Bamboo (QEMU's TCG
  mode can't emulation X2APIC)
- Crank the Bamboo untyped caps max to 230, same as simulation


AdrianDanis:
Thanks for the fix. And we are working on a better system to prevent the bamboo configs from diverging

PR #6, opened 2017-06-07
Polluted environment when running make compile-all

mokshasoft:
The environment variables from different runs affect compilation in the
compile-all target, which causes the compilation of some defconfigs to
fail. Fix by compiling each kernel_elf target in a clean environment.

### DIFF
--- a/tools/compile-all.sh
+++ b/tools/compile-all.sh
@@ -8,7 +8,7 @@ do
     make clean > /dev/null
     make $config > /dev/null
     make silentoldconfig > /dev/null
-    make kernel_elf > /dev/null
+    env -i PATH="$PATH" make kernel_elf > /dev/null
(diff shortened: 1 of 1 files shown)


AdrianDanis:
Hi @mokshasoft ,
Thanks for this and your other pull requests. Unfortunately we cannot merge changes to repositories in the [seL4](https://github.com/seL4) organization without having a [signed CLA](http://sel4.systems/Community/Contributing/).
Is this something you would be able to do?

Adrian

mokshasoft:
Hi @AdrianDanis ,
Now I have sent the signed CLA.
Jonas

AdrianDanis:
Thanks! I don't think there's any problems with this or the other changes. Just have to put them through regression before they'll merge, hopefully today or early next week.

PR #7, opened 2017-06-18
Add compile-all-clang target

mokshasoft:
The compile-all-clang target runs the compile-all.sh script with clang
as the compiler.

I am using this target to test my clang makefile porting efforts. I guess I will have a pull request soon with the changes needed to compile the kernel for most configs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,6 +33,10 @@ endif
 compile-all:
 	${PWD}/projects/sel4test/tools/compile-all.sh
 
+# test that all configs compile with clang
+compile-all-clang:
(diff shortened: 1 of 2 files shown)


AdrianDanis:
I love that you're getting clang compilation working, but I would like to hold off merging this PR until the kernel does in fact compile with clang and this target can run successfully, otherwise it may give the impression that the kernel currently supports being compiled with clang, when it does not.

pingerino:
I'm closing this, as the script has been removed and this will need to be redone for cmake.

PR #11, opened 2018-05-28
riscv: Modify BAD_ADDR to be 39-bit to pass on Sv39 systems

heshamelmatary:
Sv39 riscv only has 39-bit virtual addresses. Tests that run on Sv39 hardware fail with 48-bit addresses.

### DIFF
--- a/apps/sel4test-tests/src/tests/faults.c
+++ b/apps/sel4test-tests/src/tests/faults.c
@@ -48,8 +48,13 @@ enum {
 #define BAD_VADDR 0xf123456C
 #endif
 #elif CONFIG_WORD_SIZE == 64
+#ifdef CONFIG_ARCH_RISCV_RV64
+/* Sv39: virtual address we test is in the valid 39-bit portion of the virtual address space */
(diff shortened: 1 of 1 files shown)


AdrianDanis:
Can you provide a bit more explanation for this. Your commit message implies that the tests do not currently pass, yet we have rv64 passing in regression right now without this change, what is the difference here?

heshamelmatary:
I ran sel4test on [a board](https://www.sifive.com/documentation/freedom-soc/freedom-u500-platform-brief/) which is Sv39, and all asserts that use BAD_VADDR failed. The badaddr returned by RISC-V/seL4 is only 39-bits, which indicates that on real Sv39 hardware the CPU address port width is wired to 39 bits

AdrianDanis:
So is this an intentional vagueness in the RISC-V specification, or is one of spike or sifive acting outside of what should be correct behavior?

AdrianDanis:
Well from my point of view it would be nice if the spec explicitly stated whether the behavior is implementation dependent or not.

But for this change itself I have no problem. Just waiting to confirm CLA

PR #13, opened 2018-07-22
riscv: update sel4test for functionality on rocket-chip.

chrisguikema:
The FRAMEEXPORTS0001 test tries to allocate a 1GiB page. The
rocket-chip only has 256MiB, so the test will fail, even though
the SV39 virtual memory system supports 1GiB mappings.

The PAGEFAULT tests need the virtual address updated to be in
valid virtual memory space on the hardware.

Change-Id: I81a0902033b54528b42e4f653f6bbf1027c31218

### DIFF
--- a/apps/sel4test-tests/arch/riscv/arch_frame_type.h
+++ b/apps/sel4test-tests/arch/riscv/arch_frame_type.h
@@ -18,10 +18,10 @@
 
 /* This list must be ordered by size - highest first */
 static const frame_type_t frame_types[] = {
-#if __riscv_xlen == 64
+    /* Rocket-Chip for zedboard only has 256MiB of RAM, so we can't allocate a 1GiB page */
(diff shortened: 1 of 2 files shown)


chrisguikema:
We're going to update this request based on Hesham's kernel feedback. Just updating CONFIG_BUILD_ROCKET_CHIP to CONFIG_BUILD_ROCKET_CHIP_ZEDBOARD

chrisguikema:
Okay, so we tested on a different piece of hardware and found an address that makes more sense, and therefore isn't rocket-chip specific. We'll update the request accordingly.

chrisguikema:
We found out why the virtual address needs to change on the Sv39 model. 

```Load and store effective addresses, which are 64 bits, must have bits 63–39 all equal to bit 38, or else a page-fault exception will occur.
```

So bits 63-38 must all be zero, hence the `0x3CBA987650` value.

PR #16, opened 2019-07-29
Minor fixes

pingerino:


settings.cmake
```diff
@@ -139,7 +139,9 @@ if(NOT Sel4testAllowSettingsOverride)
     endif()
 
     if(SMP)
-        set(KernelMaxNumNodes 4 CACHE STRING "" FORCE)
+        if("${KernelMaxNumNodes}" STREQUAL "")
```
kent-mcleod: Are you able to achieve this by allowing `SMP=ON` to imply 4 nodes, and `SMP=8` imply 8 nodes and `SMP=OFF` 1 node?

pingerino: Given that CMake is embraces the idea of truthiness (e.g `SMP=1` is equivalent to `SMP=ON`) and to provide backwards compatible behaviour, I've added a new var (`NUM_NODES`) which can be used to override the default of 4. Let me know what you think.

kent-mcleod: Yea that's fine.  But you will also need to add `NUM_NODES` to easy_settings.cmake also

pingerino: cool, done

easy-settings.cmake
```diff
@@ -31,5 +31,6 @@ set(VERIFICATION OFF CACHE BOOL "Only verification friendly kernel features")
 set(BAMBOO OFF CACHE BOOL "Enable machine parseable output")
 set(DOMAINS OFF CACHE BOOL "Test multiple domains")
 set(SMP OFF CACHE BOOL "(if supported) Test SMP kernel")
+set(NUM_NODES "" CACHE BOOL "(if SMP) the number of nodes (default 4)")
```
kent-mcleod: BOOL needs to be a STRING here

PR #20, opened 2019-12-13
mcs: Update sel4test for IPC invariants

xurtis:
This change reflects the changes in sel4/sel4#172.

The following changes have been made to ensure that IPC invariants can be met:

In any case where an IPC occurs over an endpoint and the receiver has no SC, the invocation will result in donation (so that the caller cannot cause the receiver to hang).

To ensure that notifications and blocked threads only have SCs that are ready for their core (or no SC at all), only configured and ready SCs may be bound to notifications or threads.


kent-mcleod:
> This change reflects the changes in [seL4/seL4#172](https://github.com/seL4/seL4/pull/172).

Closing this as the proposal in the referenced PR did not proceed.

PR #23, opened 2020-08-07
Add tests for new MCS fault cases

xurtis:
This adds tests for the changes in https://github.com/seL4/seL4/pull/216.

# TIMEOUTFAULT0004: Fault on no donated SC

When a passive thread receives an IPC and the sender has not used seL4_Call, the sender's SC is not donated and the receiver faults to indicate that it is running without a SC.

# TIMEOUTFAULT0005: Fault on client SC removed

When a passive thread receives an IPC and the sender's SC has been removed while it is blocked, receiver faults to indicate that it is running without a SC.

# TIMEOUTFAULT0006: Fault on SC removed while running

When a thread with an SC is running and its SC is removed, that thread faults.

# TIMEOUTFAULT0007: Fault on donated SC removed while running

When a passive thread is running on a donated SC and that SC is unbound, the passive thread faults as it no longer has an SC.

# TIMEOUTFAULT0008: Fault on reply without SC return

When a client makes a call that results in a nested call and a monitor replies to the first call, the client is not returned its SC, as such it faults.

# TIMEOUTFAULT0009: Fault on donation of unconfigured SC

When a passive thread receives an IPC and the sender's SC has been replaced with an unconfigured SC while it is blocked, receiver faults to indicate that it is running with an unconfigured SC.


xurtis:
Running tests on all platforms internally with the updated kernel, will need the associated kernel changes merged first.

apps/sel4test-tests/src/tests/faults.c
```diff
@@ -1069,6 +1069,353 @@ static int test_timeout_fault_nested_servers(env_t env)
 /* this test is disabled for the same reason as TIMEOUTFAULT0002 */
 DEFINE_TEST(TIMEOUTFAULT0003, "Nested timeout fault", test_timeout_fault_nested_servers, config_set(CONFIG_KERNEL_MCS))
 
+/* Test of invalid SC and no SC faults */
```
pingerino, line 1072: Did you consider adding test cases for where each of these scenarios *doesn't* have a fault handler? we want to make sure the thread still becomes inactive and the kernel doesn't die for some reason.

xurtis: I don't know of a way to effectively check that a thread isn't still running, but can certainly check that the kernel doesn't explode (or fail any asserts).

nomadeel (approved):

PR #31, opened 2020-12-21
Fix timer requirement for scheduling context tests

alexf91:
https://github.com/seL4/sel4test/commit/4f37fac63bda3dfd21b27eec4bcdccb9f041d044 added the CONFIG_HAVE_TIMER macro, but it looks like there were some copy/paste errors for SCHED_CONTEXT_0010-0013.


kent-mcleod (approved):
Thanks a lot for this!

PR #34, opened 2021-05-12
Increase execution time for SCHED0014

xurtis:
The time provided to the tasks seems to be exhausted in some occasions before a task can yield on the ODROID-XU with the exynos5410 with debug enabled.

Doubling the budget should not affect periodic scheduling for this test.

Response time analysis for the task set give the following:

 0 - WCET = 20us, min. inter-arrival = 100us, response time = 60us
 1 - WCET = 20us, min. inter-arrival = 200us, response time = 80us
 2 - WCET = 20us, min. inter-arrival = 800us, response time = 260us

Given the response time for each task is less than the inter-arrival time and they all execute at the same priority, they should settle into a regular schedule within the hyperperiod (800us).


sylGauthier (approved):
Looks good to me

PR #35, opened 2021-05-12
Ensure server prio greater than client for timeout

xurtis:
We must ensure that during the reset of the server in the timeout test that the client is not able to preempt the server. If it does, it can perform a send before the server performs a receive. If this occurs, the server will not be blocked when the test thread removes its scheduling context and the test will hang indefinitely.


pingerino:
Looks good. 

I was looking at the test file and noticed this line should be deleted, AFAICT the test is no longer disabled: https://github.com/seL4/sel4test/blame/7f4da44f11f2b4bae5f41b24bc99f7bb7a9755eb/apps/sel4test-tests/src/tests/faults.c#L1060

xurtis:
Cool, I'll remove that comment while I'm at it.

nomadeel (approved):

PR #36, opened 2021-05-27
schedcontext: Disable SCHED_CONTEXT_0012

nomadeel:
This test was re-enabled following a fix to the test conditional guards.
However upon further investigation, the test was found to be old and was
testing an old implementation of the MCS scheduler. For now, this test
is disabled pending a decision to remove or update it.

Signed-off-by: Damon Lee <Damon.Lee@data61.csiro.au>

### DIFF
--- a/apps/sel4test-tests/src/tests/schedcontext.c
+++ b/apps/sel4test-tests/src/tests/schedcontext.c
@@ -623,8 +623,9 @@ test_revoke_reply_on_call_chain_unordered(env_t env)
 
     return sel4test_get_result();
 }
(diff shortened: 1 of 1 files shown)


pingerino:
Is it possible to fix the test rather than disable? I believe the new behaviour is that the client shouldn't return as it never gets the SC back.

nomadeel:
Yeah that's the case. I need to figure out how to test that though.

nomadeel:
Well, I figured out how to test it properly, closing this.

PR #37, opened 2021-05-28
schedcontext: Fix SCHED_CONTEXT_0012 test

nomadeel:
The SCHED_CONTEXT_0012 test was based off the old MCS implementation
where an unordered reply object revoke will still give the scheduling
context back to the original owner.

Signed-off-by: Damon Lee <Damon.Lee@data61.csiro.au>


pingerino (approved):
Looks good, assuming the test passes :)

nomadeel:
I tested it and it works with the containers, so it should also work in CI. Hopefully.

PR #38, opened 2021-06-16
riscv: disable huge page for rocketchip

ruizjuanpablo:
Rocketchip platform only has 256MB of RAM, so we can't allocate
a huge 1GB page.

Signed-off-by: Juan Pablo Ruiz <juanpablo@ssrc.tii.ae>

### DIFF
--- a/apps/sel4test-tests/arch/riscv/arch_frame_type.h
+++ b/apps/sel4test-tests/arch/riscv/arch_frame_type.h
@@ -11,9 +11,12 @@
 
 /* This list must be ordered by size - highest first */
 static const frame_type_t frame_types[] = {
-    /* Rocket-Chip for zedboard only has 256MiB of RAM, so we can't allocate a 1GiB page */
+    /* Rocket-Chip for zedboard and base Rocket-Chip only have 256MiB of RAM, 
+     * so we can't allocate a 1GiB page */
     /* Polarfire has 1GiB of memory can't allocate a 1GiB page for user space */
-#if __riscv_xlen == 64 && !defined(CONFIG_BUILD_ROCKET_CHIP_ZEDBOARD) && !defined(CONFIG_PLAT_ARIANE) &&!defined(CONFIG_PLAT_POLARFIRE)
+#if __riscv_xlen == 64 && !defined(CONFIG_BUILD_ROCKET_CHIP_ZEDBOARD) \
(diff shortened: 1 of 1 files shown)


apps/sel4test-tests/arch/riscv/arch_frame_type.h
```diff
@@ -11,9 +11,12 @@
 
 /* This list must be ordered by size - highest first */
 static const frame_type_t frame_types[] = {
-    /* Rocket-Chip for zedboard only has 256MiB of RAM, so we can't allocate a 1GiB page */
+    /* Rocket-Chip for zedboard and base Rocket-Chip only have 256MiB of RAM, 
+     * so we can't allocate a 1GiB page */
     /* Polarfire has 1GiB of memory can't allocate a 1GiB page for user space */
-#if __riscv_xlen == 64 && !defined(CONFIG_BUILD_ROCKET_CHIP_ZEDBOARD) && !defined(CONFIG_PLAT_ARIANE) &&!defined(CONFIG_PLAT_POLARFIRE)
+#if __riscv_xlen == 64 && !defined(CONFIG_BUILD_ROCKET_CHIP_ZEDBOARD) \
```
axel-h, line 17: Can't we remove `CONFIG_BUILD_ROCKET_CHIP_ZEDBOARD` then and just say this applies for all Rocket Chips? On the mailing list I have raised the question what the genera state of the Rocket Chip platform is, given that the repos on github indicate it has been abandonend.

axel-h:
Seems this could be one of the cases where sel4test should try to parse the DTB passed by the bootloader and then figure out how much memory there is. And then skip the test with an explanation that the platform configuration does not support this. Which leave the chicken-egg problem, who defines what seL4test is supposed to test or if analyzing the test output is supposed to be a post-processing step by a higher entity who should know if everything is really fine in the universe.

ruizjuanpablo:
I agree to remove CONFIG_BUILD_ROCKET_CHIP_ZEDBOARD as a condition for this test and say that CONFIG_PLAT_ROCKETCHIP applies for all rocket chip. The configuration CONFIG_BUILD_ROCKET_CHIP_ZEDBOARD was actually removed from the kernel on https://github.com/seL4/seL4/commit/8b4ed9941aba01c209e37f597db21dd6c34a49a3#diff-6633bf96e2f098e25816ecca9848a8de18e0041f704567699d2cbca62edf5058

I will close this push request and resent it with style fixes and this fix.

PR #40, opened 2021-06-23
CMake: allow skipping include of settings.cmake

axel-h:
My build system currently sets up everything and then uses `add_subdirectory()` to include sel4test. This patch allow building sel4test then. Maybe there is a smarter solution here by impersonating `settings.cmake` somehow to this does nothing?


kent-mcleod:
What is your build system setting up before including sel4test?  Could you instead include `add_subdirectory(apps/sel4test-driver)` instead of the toplevel CMakeLists.txt that currently expects to setup the elfloader and kernel.elf targets?

axel-h:
Our build system basically does what `settings.cmake` does and that conflicts. The goal is being able to use existing seL4 projects like this without modification, so we don't have to create another wrapper project. Having the switch `SEL4TEST_DONT_USE_SEL4_DEFAULT_BUILD_SYSTEM` nicely solves this.

kent-mcleod:
> The goal is being able to use existing seL4 projects like this without modification, so we don't have to create another wrapper project.

But isn't this already a modification?  The coupling between settings.cmake and CMakeLists.txt isn't a stable public interface.  settings.cmake contains many sel4test specific options and options can be moved between CMakeLists.txt and settings.cmake without causing external breakages.

`add_subdirectory(apps/sel4test-driver)` or even using https://cmake.org/cmake/help/latest/variable/CMAKE_PROJECT_PROJECT-NAME_INCLUDE_BEFORE.html#variable:CMAKE_PROJECT_%3CPROJECT-NAME%3E_INCLUDE_BEFORE will let you include the sel4test applications into another project without requiring special modifications upstream.

axel-h:
Thanks, I will give this a try.

CMakeLists.txt
```diff
@@ -6,7 +6,10 @@
 
 cmake_minimum_required(VERSION 3.7.2)
 
-include(settings.cmake)
+if(NOT SEL4TEST_DONT_USE_SEL4_DEFAULT_BUILD_SYSTEM)
+    # include the settings from the standard sel4 build system
+    include(settings.cmake)
```
kent-mcleod, line 11: What if you added a variable `SEL4TEST_SETTINGS_CONFIGURED` and used it like an include guard at the top of settings.cmake
```
get_property(include_guard GLOBAL PROPERTY SEL4TEST_SETTINGS_CONFIGURED)
if (include_guard)
    return()
endif()
set_property(GLOBAL PROPERTY SEL4TEST_SETTINGS_CONFIGURED "ON")
```

axel-h: What I consider a bit ugly about this is, that all I really want to is not including `settings.cmake`. Clearly doing this in the `CMakeLists.txt` seems the cleanest thing. ... but actually, having some parts of `settings.cmake` would be nice because there are project specific settings there.I have to try this out and see where the real (build system) conflicts are with this file.

kent-mcleod: Do you still need to not include settings.cmake?  It's now safe to call sel4_configure_platform_settings() multiple times which it may not have been when this PR was originally created.

kent-mcleod (approved):

kent-mcleod:
@axel-h, feel free to merge this if you still need the change

lsf37 (approved):

axel-h:
I have not found a better solution yet for our use case with a different/custom build system. I will check our internal CI one more tomorrow and merge this then. Thanks.

Indanz:
I think this can be merged.

CMakeLists.txt
```diff
@@ -6,7 +6,10 @@
 
 cmake_minimum_required(VERSION 3.7.2)
 
-include(settings.cmake)
+if(NOT SEL4TEST_DONT_USE_SETTINGS_CMAKE)
```
axel-h, line 9: The other option I was considering is having this

```
# By default, settings.cmake is used. This can be customized:
# - use custom settings:      cmake ... -D SEL4TEST_SETTINGS_CMAKE=my-settings.cmake
# - use no settings at all:   cmake ... -D SEL4TEST_SETTINGS_CMAKE=""
if(NOT DEFINED SEL4TEST_SETTINGS_CMAKE)
    include("settings.cmake")
elseif(SEL4TEST_SETTINGS_CMAKE)
    include("${SEL4TEST_SETTINGS_CMAKE}")
endif()
```

But it turned out I never used a custom settings file anywhere, instead all custom settings were either passed on the cmake command line of by a cmake wrapper. Thus I have some doubt this is really useful.

Indanz: Do you want this PR to be merged or not? I'm confused now.

axel-h: Currently I have no immediate need for that. For the build system we had at Hensoldt Cyber, this patch used to be necessary. Not sure about the need nowadays.

lsf37:
@axel-h this PR should either be closed or merged — happy to merge it if it’s useful for you, otherwise I’d close it.

PR #45, opened 2021-08-13
CACHEFLUSH: Add test for flushing when retyping

kent-mcleod:
Reusing a frame object after revoking it and retyping it again shouldn't
see any old data from before the retype. The kernel should ensure that
any clear memory operations have been flushed completely out so that any
reads that bypass the caching hierarchy can't see old data.

Signed-off-by: Kent McLeod <kent@kry10.com>

### DIFF
--- a/apps/sel4test-tests/src/tests/cache.c
+++ b/apps/sel4test-tests/src/tests/cache.c
@@ -11,6 +11,7 @@
 #include <stdio.h>
 #include <sel4/sel4.h>
 #include <vka/object.h>
+#include <vka/capops.h>
 #include <sel4utils/util.h>
 #include <sel4utils/arch/cache.h>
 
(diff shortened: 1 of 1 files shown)


lsf37 (approved):

apps/sel4test-tests/src/tests/cache.c
```diff
@@ -253,3 +254,78 @@ DEFINE_TEST(CACHEFLUSH0003, "Test that cache maintenance can be done on large pa
...
+    /* Revoke the untyped. This deletes the frame. */
+    vka_cnode_revoke(&src_path);
+
+    /* Create the frame with the same untyped. The kernel guarantees that the contents have been zeroed. */
+    error = seL4_Untyped_Retype(untyped, seL4_ARCH_4KPage, PAGE_BITS_4K, dest_path.root, dest_path.dest,
+                                dest_path.destDepth, dest_path.offset, 1);
+    test_error_eq(error, seL4_NoError);
+
+
+    /* map in the frame without cacheability again. */
```
axel-h, line 320: I think this test should have a bit more explanation what is tested here and what not. If we map a frame uncached, revoke it and map it again uncached, the accesses always bypasses the cache and is supposed to go directly to memory. So all the test is dosing is checking that the kernel clears the content. What is does not test here is that proper cache management also happens, right? So we still need a test that maps a page cached, populated it with markers, then ensure this is written to RAM (this seems the tricky part). Then revoke the frame, map it again, but this time uncached. And then the whole frame should be empty, so we can be sure the page cleaning has not just happened in the cache, but it did not make it into RAM yet as no eviction has happened.

lsf37: Would not hurt to add that case for completeness, yes.

There are more, actually, there are also kernel objects, and as @kent-mcleod figured out last night also the i-cache. So ideally we'd add additional tests for:

1. cached framed with content -> retype -> uncached frame
2. other kernel objects with content -> retype -> uncached frame
3. any memory/cache content (any of the combinations from above) -> retype -> uncached frame mapped execute and trying to infer information from the i-cache

There's no really reliable way to make sure test 1 is written to RAM unless you manually flush it, which we could do for the test.

Smattr: It's possible to have a frame mapped in twice, once cached and once uncached. To test you hit RAM, you can write through the cached mapping and read back through the uncached mapping. Though this doesn't work on some of the common architectures where reads through uncached mappings still snoop the cache.

kent-mcleod: > So all the test is dosing is checking that the kernel clears the content. What is does not test here is that proper cache management also happens, right?

The kernel doesn't clear the content through the same mapping, but through it's own separate mapping that is cached.  This test is checking that the clearing by the kernel ends up as ram.  It is checking that the proper cache management is occurring (for cleaning).  It doesn't check that other caches are invalidated.  If a future change was made such that the kernel reset memory via uncached mappings and didn't invalidate the caches, then this test wouldn't detect that a frame mapped cached after a retype could also see stale data.

Issue #42, opened 2021-07-17
Test SCHED0021 has a race condition


axel-h:
```
    /* Configure all of the threads */
    for (size_t thread = 0; thread < PREEMPTION_THREADS; thread += 1) {
[...]
        /* Thread must run at same prio as monitor */
        set_helper_priority(env, &threads[thread], OUR_PRIO);
[...]
    }
[...]
    /* Start executing other threads */
    ZF_LOGD("Releasing Threads");
    test_simple_preempt_start = 1;
    /* Yield should cause all other threads to execute before returning
     * to the current thread. */
    seL4_Yield();
```
Isn't there a tiny chance that the preemptive scheduler will kick in after `test_simple_preempt_start = 1`? So all threads get their time slice to run. Then `seL4_Yield()` is executed and each thread again gets a time slice to run? 
Wouldn't is be better to do this:
```
    /* Start executing other threads */
    ZF_LOGD("Releasing Threads");
    /* Release our time slice now to get a new one. That should ensure we are
     * not interrupted due to bad luck by the preemptive scheduler right after 
     * setting test_simple_preempt_start to 1 and then voluntarily yielding. For
     * the same reason we are also printing the doing the ZF_LOGD() above 
     * already.
     */
    seL4_Yield();
    /* Set a timeout for the test. Each thread should run for one tick. */
    uint64_t start = time_now(env);
    test_simple_preempt_start = 1;
    /* Yield should cause all other threads to execute before returning
     * to the current thread. 
     */
    seL4_Yield();
    test_simple_preempt_start = 0;
```
There is still a risk because we can't be really sure what `time_now(env)` does and how long this may take. 

Aside from this, the line 
```
    uint64_t now = start;
```
is the original code seems quite useless to me, as nobody is using the value assigned to `now`, it is overwritten a few lines later. This might be a left over from a time where another debug message was printed with it?

lsf37:
After a long time of silence there has been another occurrence of SCHED0021 failing here:

https://github.com/seL4/seL4/actions/runs/29725883787#pc99-debug-gcc-32-sched0021

```
<testcase classname="sel4test" name="SCHED0021">
<error>Check duration(757907247) &lt; ((((2 * 5 * (1000ULL * 1000ULL)) * (4 + 1)) * (102)) / (100))(51000000) failed. at line 1635 of file /github/workspace/projects/sel4test/apps/sel4test-tests/src/tests/scheduler.c</error>
[...]

PR #46, opened 2021-08-21
switch off tests that are flaky under simulation

lsf37:
These are currently BREAKPOINT_002 and SCHED0021. The corresponding issues #43 and #42 remain open, and when resolved positively, these tests should be enabled again for simulation runs.

### DIFF
--- a/apps/sel4test-tests/src/tests/breakpoints.c
+++ b/apps/sel4test-tests/src/tests/breakpoints.c
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: BSD-2-Clause
  */
 #include <autoconf.h>
+#include <sel4test-driver/gen_config.h>
 
 #ifdef CONFIG_HARDWARE_DEBUG_API
 
@@ -263,8 +264,11 @@ test_debug_set_data_breakpoint(struct env *env)
     }
     return sel4test_get_result();
 }
+/* This test is flaky under simulation. See also #43  */
 DEFINE_TEST(BREAKPOINT_002, "Attempt to set and trigger a data breakpoint",
-            test_debug_set_data_breakpoint, config_set(CONFIG_HARDWARE_DEBUG_API))
+            test_debug_set_data_breakpoint,
+            config_set(CONFIG_HARDWARE_DEBUG_API) &&
+            !config_set(CONFIG_SIMULATION))
 
 static int
 test_debug_get_instruction_breakpoint(struct env *env)
--- a/apps/sel4test-tests/src/tests/scheduler.c
+++ b/apps/sel4test-tests/src/tests/scheduler.c
@@ -1624,8 +1624,10 @@ static int test_simple_preempt(struct env *env)
 
     return sel4test_get_result();
 }
+/* This test is flaky under simulation. Probably a race condition that only
+   comes out under simulator timing conditions. See also #42  */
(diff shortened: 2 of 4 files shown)


apps/sel4test-tests/src/tests/breakpoints.c
```diff
@@ -264,7 +265,9 @@ test_debug_set_data_breakpoint(struct env *env)
     return sel4test_get_result();
 }
 DEFINE_TEST(BREAKPOINT_002, "Attempt to set and trigger a data breakpoint",
-            test_debug_set_data_breakpoint, config_set(CONFIG_HARDWARE_DEBUG_API))
+            test_debug_set_data_breakpoint,
+            config_set(CONFIG_HARDWARE_DEBUG_API) &&
+            !config_set(CONFIG_SIMULATION))
```
axel-h, line 270: Can you add a comment here why this test does not work ion the simulation? I assume QEMU jst does not support this?

apps/sel4test-tests/src/tests/scheduler.c
```diff
@@ -1625,7 +1625,7 @@ static int test_simple_preempt(struct env *env)
     return sel4test_get_result();
 }
 DEFINE_TEST(SCHED0021, "Test for pre-emption during running of many threads with equal prio", test_simple_preempt,
-            true);
+            !config_set(CONFIG_SIMULATION));
```
axel-h, line 1628: Same here, should have a comment why this does not work in the simulation. We could mention https://github.com/seL4/sel4test/issues/42 here.

apps/sel4test-tests/src/tests/breakpoints.c
lsf37: Good point, will add a comment. I've only observed the problem in #43 under simulation so far. If it really is a race condition, it seems to be only triggered under simulation conditions.

If we see it elsewhere as well then we should definitely change the flag, but I haven't yet so far. In fact, the only flaky test I've seen so far on hardware is `MULTICORE0004`, which does say `(flaky)` in its description.

What is flaky in the hardware tests is mostly boards not coming up for boot or not recognising correctly that they have booted and then timing out on the boot process. Problems ranging from "network cable fell out", to dodgy earth connection messing up serial, to old cable etc. Will hopefully become stable again over time.

kent-mcleod (approved):

apps/sel4test-tests/src/tests/breakpoints.c
```diff
@@ -263,8 +264,11 @@ test_debug_set_data_breakpoint(struct env *env)
     }
     return sel4test_get_result();
 }
+/* This test is flaky under simulation. See also #43  */
```
axel-h, line 267: should better day "GitHub Issue #43" to give a bit more context that this numer refers to.

apps/sel4test-tests/src/tests/scheduler.c
```diff
@@ -1624,8 +1624,10 @@ static int test_simple_preempt(struct env *env)
 
     return sel4test_get_result();
 }
+/* This test is flaky under simulation. Probably a race condition that only
+   comes out under simulator timing conditions. See also #42  */
```
axel-h, line 1628: Same here, better say "GitHub Issue #42" to give a bit more context that this number refers to.

PR #49, opened 2021-08-28
add missing return statement (fix #47)

### LINKED ISSUE #47: sched0022_to_fn() does not return a value
The function `sched0022_to_fn()` does not return a value and the compiler complains
```
  [249/267] Building C object .../src/tests/scheduler.c.obj
  .../src/tests/scheduler.c:1653:1: warning: control reaches end of non-void function [-Wreturn-type]
  }
  ^
  1 warning generated.
```

I wonder, is there just a `return sel4test_get_result();` missing?

axel-h:
add missing return statement as fix https://github.com/seL4/sel4test/issues/47 for. @yyshen suggested just to return 0 here, but this seem to contradict what all other test are doing.

### DIFF
--- a/apps/sel4test-tests/src/tests/scheduler.c
+++ b/apps/sel4test-tests/src/tests/scheduler.c
@@ -1652,6 +1652,7 @@ int sched0022_to_fn(struct env *env, helper_thread_t *thread, seL4_CPtr ep)
 
     seL4_SetMR(1, error);
     seL4_Send(ep, tag);
+    return 0;
 }
 
 /* Test that a helper thread can move itself back from another core.


lsf37:
The function `sched0022_to_fn()` is not itself a test, it's only a helper function inside a test. There are plenty of these helper functions that do `return 0`, e.g. in the same file `prio_test_func`:

https://github.com/seL4/sel4test/blob/a9d3d683c04ecab5a1a4f4b00238044f24c72b24/apps/sel4test-tests/src/tests/scheduler.c#L252-L266

I think `return 0` is right here. `sel4test_get_result()` will probably not do anything bad, because the return value of `sched0022_to_fn()` is ignored anyway in this test, but it's not what is expected at that point.

axel-h:
change.

lsf37 (approved):

PR #50, opened 2021-09-03
Allow Lazy SC Rebind

Indanz:
It is now allowed to bind a scheduling context to both a notification
object as well as a TCB. When that is the case, the SC will be unbound
from the TCB when the task blocks on that notification, making the task
passive. Just as if the task has been running on a SC bound to the
notification.

Signed-off-by: Indan Zupancic <Indan.Zupancic@mep-info.com>

[Edit for test:
Test with seL4/seL4#523 ]


Indanz:
This is the sel4test counterpart of seL4/seL4/#523.

lsf37 (approved):

PR #56, opened 2021-11-02
Add test for race in seL4 issue 633

Indanz:
This triggers a race between scheduling context deletion and the out
of budget timer IRQ. This test is very timing sensitive, it is known
to trigger the race condition on tqma8xqp1gb with RELEASE disabled.

Also add MIN_BUDGET tests. This assumes WCET is at most 10us on
all tested platforms!

### DIFF
--- a/apps/sel4test-tests/src/tests/schedcontext.c
+++ b/apps/sel4test-tests/src/tests/schedcontext.c
@@ -12,6 +12,10 @@
 
 #include "../helpers.h"
 
+#ifndef MIN_BUDGET_US
+#define MIN_BUDGET_US (2 * 10)
+#endif
+
 int test_sched_control_configure(env_t env)
 {
     int error;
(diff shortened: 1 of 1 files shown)


Indanz:
This depends on https://github.com/seL4/seL4/pull/634 being merged, otherwise it will trigger an assertion.

Indanz:
This should be ran on real hardware before being merged. Hopefully it triggers the same bug on another platform than tqma8xqp1gb, so we have an extra way to test whatever fix we do.

lsf37:
Cool, thanks for this. I'll see if I can find a platform that triggers this.

lsf37:
Interestingly ia32 triggers the assertion failure, but none of the other Arm platforms in the test so far (tx2 is still running). But that is good enough to confirm its absence after the fix and to catch it should it re-occur.

apps/sel4test-tests/src/tests/schedcontext.c
```diff
@@ -46,6 +46,14 @@ int test_sched_control_configure(env_t env)
     error = api_sched_ctrl_configure(simple_get_sched_ctrl(&env->simple, 0), sc, 5000llu, 1000llu, 0, 0);
     test_eq(error, seL4_RangeError);
 
+    /* test budget < MIN_BUDGET doesn't work */
+    error = api_sched_ctrl_configure(simple_get_sched_ctrl(&env->simple, 0), sc, 1, 5000, 0, 0);
```
lsf37, line 50: Not that it has very different semantics in this case, but the other API calls use `5000llu` instead of `5000` etc. It'd be nice to keep those consistent.

apps/sel4test-tests/src/tests/schedcontext.c
```diff
@@ -46,6 +46,14 @@ int test_sched_control_configure(env_t env)
...
+    test_eq(error, seL4_RangeError);
+
+    /* test budget == MIN_BUDGET does work */
+    error = api_sched_ctrl_configure(simple_get_sched_ctrl(&env->simple, 0), sc, 2 * 10, 5000, 0, 0);
```
lsf37, line 54: This should probably use a named constant/macro instead of `10`. Kernel WCET tends to be set to `10` for most platforms, but not for all of them, and it's a bit of a magic number here. @kent-mcleod is some form of the name `MIN_BUDGET` available in this scope?

lsf37 (approved):

apps/sel4test-tests/src/tests/schedcontext.c
kent-mcleod: No there isn't.  Both kernel constants MIN_BUDGET_US and MAX_PERIOD_US should be available as constants in a libsel4 header somewhere but currently aren't.  Maybe for now it'd be fine to just define MIN_BUDGET_US in this file as a constant and then update it once we add the constants to libsel4?

apps/sel4test-tests/src/tests/schedcontext.c
Indanz: That's what I started with, but then I ended up with ``1llu`` and that's just too ugly. I can't think of a good reason why llu would improve anything other than consistency, it just makes things less readable. Okay if I remove the others instead?

apps/sel4test-tests/src/tests/schedcontext.c
Indanz: I added
```
#ifndef MIN_BUDGET_US
#define MIN_BUDGET_US (2 * 10)
#endif
```

apps/sel4test-tests/src/tests/schedcontext.c
```diff
@@ -12,6 +12,10 @@
 
 #include "../helpers.h"
 
+#ifndef MIN_BUDGET_US
+#define MIN_BUDGET_US (2 * 10)
+#endif
```
lsf37, line 17: So, I merged this a bit too early -- this breaks for TK1 where KERNEL_WCET is set to 100 instead of 10

Indanz: Best fix would be to export MIN_BUDGET_US in libsel4.

Indanz: Alternatively, stop the tests without error if the api_sched_ctrl_configure() call fails.

Indanz: I couldn't figure out how to export MIN_BUDGET_US from seL4, so instead changed the hardcoded MIN_BUDGET_US define to 100 for TK1. Not sure if CONFIG_PLAT_TK1 is set, but that's what others are using. See commit b2c24b0.

lsf37: Thanks, I added a comment and made that into a PR. Exporting MIN_BUDGET_US is indeed a bit annoying because drags a few other things with it.

Issue #61, opened 2021-12-07
SCHED0000 fails for PC99_release with clang 11


lsf37:
The failure happens only when compiled with clang 11 (gcc 10 is fine), and only in release configs (works for debug), and only for PC99 (both 32 and 64 bit, Arm and RISCV are fine).

The test fails *only* under simulation/qemu. It passes fine for all configs on hardware.

```xml
<testcase classname="sel4test" name="SCHED0000">
  <failure type="failure">result == SUCCESS at line 291 of file /workspace/projects/sel4test/apps/sel4test-driver/src/testtypes
  <error>result == SUCCESS at line 217 of file /workspace/projects/sel4test/apps/sel4test-driver/src/main.c</error>
</testcase>
```

This is the last remaining known blocker for seL4/seL4-CAmkES-L4v-dockerfiles#39

PR #62, opened 2021-12-07
Resolve issues with SCHED0000 test

kent-mcleod:
Resolves https://github.com/seL4/sel4test/issues/61


lsf37 (approved):
Nice! This should unblock the new docker containers as well.

PR #63, opened 2021-12-08
scheduler: Increase timeout period when simulating

kent-mcleod:
Waiting for two timeouts is not enough when simulating in cloud
instances. Increase the timeout period under simulation to 100ms.

Signed-off-by: Kent McLeod <kent@kry10.com>


lsf37:
Still seems to be failing unfortunately, this time for both 32 and 64 bit: https://github.com/seL4/sel4test/runs/4464611120?check_suite_focus=true#step:3:8745

lsf37 (approved):
That seems to have done it, now passing consistently.

PR #74, opened 2022-03-10
cmake: create script only if SIMULATION is enabled

ichdream:
The simulation scripts are useless if the target is a real board and
not a QEMU simulation. Thus they will only be created if SIMULATION
is explicitly ON (-DSIMULATION=TRUE).
SIMULATION is disabled by default

Signed-off-by: ichdream <chensheng0802@outlook.com>


axel-h (approved):
Looks good to me. Basically all from https://github.com/seL4/sel4test/pull/73 in a clean PR.

ichdream:
> Looks good to me. Basically all from #73 in a clean PR.

Thanks for your advice, it looks clearer now.

PR #76, opened 2022-06-02
MCS: Add periodic drift test

Indanz:
If this test fails, TIMER_OVERHEAD_TICKS may not be set correctly
for the platform. Needs seL4 pull #847 applied to pass.

Signed-off-by: Indan Zupancic <Indan.Zupancic@mep-info.com>


Indanz:
Without https://github.com/seL4/seL4/pull/847 this fails with 2498 and 2494 for SMP instead of the expected 2500 on tqma8xqp1gb.

kent-mcleod (approved):
If I understand this right, the main thread sleeps for (1 second - half-a-period  = 1 - 0.000002s) which should guarantee that each of the child threads should be allowed to yield exactly (1 second / period = 2500) times.  This means that if a period drifts by more than 80ns for each activation on average the test will fail, but if it's less than 80ns it will pass?

Indanz:
Correct. The test can be made less strict by reducing the sleep time.

apps/sel4test-tests/src/tests/scheduler.c
```diff
@@ -1697,3 +1697,65 @@ static int test_changing_affinity(struct env *env)
...
+}
+DEFINE_TEST(SCHED0023, "Test period drift", test_period_drift,
+            config_set(CONFIG_KERNEL_MCS) &&config_set(CONFIG_HAVE_TIMER))
+
+int test_period_drift_smp(env_t env)
+{
+    return run_test_period_drift(env, 2);
+}
+DEFINE_TEST(SCHED0024, "Test period drift SMP", test_period_drift_smp,
+            config_set(CONFIG_KERNEL_MCS) &&config_set(CONFIG_HAVE_TIMER) &&CONFIG_MAX_NUM_NODES > 1)
```
axel-h, line 1761: nitpicking:
```suggestion
            config_set(CONFIG_KERNEL_MCS) && config_set(CONFIG_HAVE_TIMER) && CONFIG_MAX_NUM_NODES > 1)
```

apps/sel4test-tests/src/tests/scheduler.c
```diff
@@ -1697,3 +1697,65 @@ static int test_changing_affinity(struct env *env)
...
+static int run_test_period_drift(env_t env, int num_threads)
+{
+    int error;
+    helper_thread_t t[num_threads];
+    volatile unsigned long counters[num_threads];
+
+    /* Create tasks with lower priority: */
+    for (int i = 0; i < num_threads; i++) {
+        create_helper_thread(env, &t[i]);
+        set_helper_priority(env, &t[i], env->priority - 1);
```
axel-h, line 1723: maybe add an assert here that this does not cause a roll over.

axel-h:
> If I understand this right, the main thread sleeps for (1 second - half-a-period = 1 - 0.000002s) which should guarantee that each of the child threads should be allowed to yield exactly (1 second / period = 2500) times. This means that if a period drifts by more than 80ns for each activation on average the test will fail, but if it's less than 80ns it will pass?

@Indanz Could you add this as a comment that describes the test's details?

apps/sel4test-tests/src/tests/scheduler.c
Indanz: That's what I had, but that is disallowed by the style checker. :-(

apps/sel4test-tests/src/tests/scheduler.c
Indanz: Good idea. `test_timeout_fault_in_server` does `OUR_PRIO - 2`, and `#define OUR_PRIO (env->priority)`. It seems priority is set to `seL4_MaxPrio - 1` in main.c.

apps/sel4test-tests/src/tests/scheduler.c
lsf37: There is something strange going on with `&&` in the style checker. I've seen the same kind of behaviour in another place recently and it looks like a bug to me (it seems to be parsing it as an address operator, at least that would be consistent with what it's printing out). You could try to spread it over multiple lines. Not sure if that will placate it, but I think that would still be better than the mushed-together form it is forcing right now.

PR #85, opened 2022-12-08
Add tests for SMC Capability

Furao:
This new test will make sure that the SMC capability and its associated API call work as expected.

It tests the following:
1. A correctly configured call works
2. An SMC call with a function ID that uses a cap with a different badge fails
3. Minting an SMC cap from an SMC cap with a badge other than 0 fails

I tested this with the following configuration:
```
../init-build.sh -DPLATFORM=zcu102 -DRELEASE=true -DARM_HYP=TRUE -DKernelAllowSMCCalls=true
```
I wanted to test on QEMU, but it turned out to be non-trivial to integrate ATF with QEMU for sel4Test. It is doable, but I don't have the bandwidth to support it.

Note: This test assumes you are running ATF in EL3 on an AARCH64 platform. The successful test checks the standard service version and it reports back 0.1 (See https://github.com/ARM-software/arm-trusted-firmware/blob/master/include/services/std_svc.h). Technically this could change in the future and break the test, however the version hasn't changed in the last 9 years since the service was written. 

This test requires these PRs:
- https://github.com/seL4/seL4_projects_libs/pull/24
- https://github.com/seL4/seL4/pull/701

### DIFF
--- a/apps/sel4test-tests/src/tests/smc.c
+++ b/apps/sel4test-tests/src/tests/smc.c
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2017, Data61, CSIRO (ABN 41 687 119 230)
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include <autoconf.h>
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <sel4/sel4.h>
+#include <sel4utils/arch/util.h>
+
+#include <vka/object.h>
+#include <vka/capops.h>
+
+#include "../test.h"
+#include "../helpers.h"
+#include "sel4/simple_types.h"
+
+#ifdef CONFIG_ALLOW_SMC_CALLS
+
+#define ARM_STD_SVC_VERSION    0x8400ff03
+#define UNASSIGNED_SMC         0x82000000
+
+static int test_smc_calls(env_t env)
+{
+    seL4_ARM_SMCContext smc_args;
+    seL4_ARM_SMCContext smc_results;
+    int error;
+
+    seL4_CPtr smc_cap = env->smc;
+    seL4_CPtr badged_smc_cap = get_free_slot(env);
+
+    error = cnode_mint(env, smc_cap, badged_smc_cap, seL4_AllRights, ARM_STD_SVC_VERSION);
+    test_error_eq(error, seL4_NoError);
+
+    /* Set function and arguments */
+    smc_args.x0 = ARM_STD_SVC_VERSION;
+    smc_args.x1 = 0;
+    smc_args.x2 = 2;
+    smc_args.x3 = 3;
+    smc_args.x4 = 0;
+    smc_args.x5 = 0;
+    smc_args.x6 = 0;
+    smc_args.x7 = 0;
+
+    /* This should succeed */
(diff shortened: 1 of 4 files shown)


apps/sel4test-tests/src/tests/smc.c
```diff
@@ -0,0 +1,70 @@
...
+    smc_args.x7 = 0;
+
+    /* This should succeed */
+    error = seL4_ARM_SMC_Call(badged_smc_cap, &smc_args, &smc_results);
+    test_error_eq(error, seL4_NoError);
+
+    test_eq(smc_results.x0, 0UL);
+    test_eq(smc_results.x1, 1UL);
+    test_eq(smc_results.x2, 0UL);
+    test_eq(smc_results.x3, 0UL);
```
Indanz, line 54: Results are put in x0 and x1, so perhaps initialise x2 and x3 to 2 and 3, to check that the kernel doesn't clobber anything? And perhaps make it future proof by checking that (major + minor) >= 0.1?

Furao: Good ideas. I will do that. Although, major + minor would be 1, so it should be (major + minor) >= 1, correct?

Furao: Although, that adjustment would pass if `seL4_ARM_SMC_Call()` did absolutely nothing. So I think I should also add a check to make sure that `smc_results.x0 != ARM_STD_SVC_VERSION`.

Furao: > Results are put in x0 and x1, so perhaps initialize x2 and x3 to 2 and 3, to check that the kernel doesn't clobber anything?

I tried this, but the SMC call clobbers it. 

image

We can't actually trust anything in the `seL4_ARM_SMCContext` to be unmodified.

apps/sel4test-tests/src/tests/smc.c
```diff
@@ -0,0 +1,77 @@
...
+    test_error_eq(error, seL4_IllegalOperation);
+
+    /* This should fail - can't mint from cap with non-zero badge */
+    seL4_CPtr bad_badged_cap = get_free_slot(env);
+    error = cnode_mint(env, badged_smc_cap, bad_badged_cap, seL4_AllRights, UNASSIGNED_SMC);
+    test_error_eq(error, seL4_IllegalOperation);
+
+    return sel4test_get_result();
+}
+DEFINE_TEST(SMC0001, "Test SMC calls", test_smc_calls, true)
```
kent-mcleod, line 75: Is it reasonable to assume that `CONFIG_ALLOW_SMC_CALLS` will only be enabled on sel4test runs when the test platform also has a compatible ATF running for this test to pass?

kent-mcleod: (It seems reasonable to me)

kent-mcleod (approved):

apps/sel4test-tests/src/tests/smc.c
Furao: I think so too. You really need to go looking to add it in.

Indanz: I thought ARM mandates ATF for Aarch64, or at least strongly encourages it.

Indanz (approved):

PR #86, opened 2023-01-18
better defaults for tests with NUM_DOMAINS > 1

lsf37:
We are currently not running any tests with NUM_DOMAINS > 1 and the defaults in sel4test lead to extremely slow test execution.

This PR picks nicer defaults:

- only 4 domains instead of 16
- give domain 0 a long time slice
- some schedule variety in the first 4 domains
- shortest period of 2 ticks

This should enable faster tests even under simulation.

This PR is a prerequisite for seL4/ci-actions#248.

### DIFF
--- a/domain_schedule.c
+++ b/domain_schedule.c
@@ -6,7 +6,12 @@
 
 /* This is a domain schedule that is suitable for the domains tests in sel4test. All
  * sel4test actually needs is for every domain to be executable for some period of time
- * in order for the tests to make progress
+ * in order for the tests to make progress.
+ *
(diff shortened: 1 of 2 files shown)


lsf37:
@robs-cse this should be close to the values you used in your test run. Does this look overall reasonable?

Xaphiosis (approved):

robs-cse (approved):
These defaults look very reasonable to me and sel4test works fine when I run it with this domain schedule (with KernelNumDomains set to either of 4 or 16) on `qemu-riscv-virt`.

kent-mcleod (approved):

PR #89, opened 2023-03-09
use config_set only for boolean config values

lsf37:
config_set is true for 1 and false for all other values, so it does not make sense to use with CONFIG_MAX_NUM_NODES.

The two instances where this happened look like they are trying to guard CONFIG_MAX_NUM_NODES > 1 by checking whether CONFIG_MAX_NUM_NODES has any value at all, but doing so incorrectly. CONFIG_MAX_NUM_NODES has a default value of 1 and should always be set to a number, so we can just drop the guard. If this assumption about always having a numbers value is wrong, the code will not compile, which is acceptable.

Thanks to @jearc for spotting these.

This wrong setting has masked two SMP tests, which may now be failing.

### DIFF
--- a/apps/sel4test-tests/src/tests/fpu.c
+++ b/apps/sel4test-tests/src/tests/fpu.c
@@ -187,4 +187,4 @@ int smp_test_fpu(env_t env)
     return sel4test_get_result();
 }
 DEFINE_TEST(FPU0002, "Test FPU remain valid across core migration", smp_test_fpu,
-            config_set(CONFIG_MAX_NUM_NODES) &&config_set(CONFIG_HAVE_TIMER) &&CONFIG_MAX_NUM_NODES > 1)
+            config_set(CONFIG_HAVE_TIMER) &&(CONFIG_MAX_NUM_NODES > 1))
--- a/apps/sel4test-tests/src/tests/ipc.c
+++ b/apps/sel4test-tests/src/tests/ipc.c
@@ -1361,6 +1361,8 @@ static int test_sched_donation_cross_core(env_t env)
 
     return sel4test_get_result();
 }
-DEFINE_TEST(IPC0028, "Cross core sched donation", test_sched_donation_cross_core,
-            config_set(CONFIG_KERNEL_MCS) &&(CONFIG_MAX_NUM_NODES > 1));
+/* This test currently fails.
+   See https://github.com/seL4/seL4/issues/941 and https://github.com/seL4/seL4/pull/986 */
+DEFINE_TEST(IPC0028, "Cross core sched donation", test_sched_donation_cross_core, false);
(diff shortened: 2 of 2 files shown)


axel-h:
Could this be something from lone time ago when SMP support was added. So checking `CONFIG_MAX_NUM_NODES' implied checking SMP support?

lsf37:
> Could this be something from lone time ago when SMP support was added. So checking `CONFIG_MAX_NUM_NODES' implied checking SMP support?

It's possible that this is very old and made sense in a previous build system. There was a previous commit (1612acca477697f) that fixed a lot of these, but seems to have missed these two.

lsf37:
I should note that I have checked all other instances of `config_set` now and they all refer to boolean values only, so these should really be the last ones unless somebody introduces new instances.

apps/sel4test-tests/src/tests/fpu.c
```diff
@@ -187,4 +187,4 @@ int smp_test_fpu(env_t env)
     return sel4test_get_result();
 }
 DEFINE_TEST(FPU0002, "Test FPU remain valid across core migration", smp_test_fpu,
-            config_set(CONFIG_MAX_NUM_NODES) &&config_set(CONFIG_HAVE_TIMER) &&CONFIG_MAX_NUM_NODES > 1)
+            config_set(CONFIG_HAVE_TIMER) &&CONFIG_MAX_NUM_NODES > 1)
```
axel-h, line 190: maybe improve the readability also 
```suggestion
            config_set(CONFIG_HAVE_TIMER) && (CONFIG_MAX_NUM_NODES > 1))
```

axel-h (commented):
Looks ok to me.

lsf37:
The `FPU0002 ` test seems to be passing, but as expected `IPC0028` is failing.

apps/sel4test-tests/src/tests/fpu.c
lsf37: Let's try that and see if the style checker accepts it

lsf37: Unforunately (and annoyingly) it insists on removing the space after `&&`, but I think the parentheses do help in this case, so I'll add them.

kent-mcleod (approved):

lsf37:
We should only merge this one after a fix for `IPC0028` is merged (e.g. something like https://github.com/seL4/seL4/pull/986).

Indanz (approved):

axel-h (approved):

lsf37:
The offending commit that introduced these was https://github.com/seL4/sel4test/commit/d07b2383b563015d8e7dad5c587493be01cf8b3c from 2017. I've now added a commit on top that disables IPC0028 so we can at least re-enable the other test.

I'll make a new PR that re-enables IPC0028 which we can use for testing possible solutions for the issue with IPC0028.

PR #92, opened 2023-05-08
aarch64: Use 32-bit sized register names

kent-mcleod:
The fault assembly code manipulates an integer sized argument and so 32-bit register names need to be used in order to avoid compiler warnings.


axel-h:
Why not simply change the type from int to word?

kent-mcleod:
> Why not simply change the type from int to word?

Cause x86_64 is already using 32-bit register names

axel-h (approved):

PR #94, opened 2023-05-23
Add tests for all of the ARM cache operations on all kinds of frame mappings.

Ivan-Velickovic:
This PR aims to add test coverage for https://github.com/seL4/seL4/issues/214. More specifically. it implement tests for the scenarios outlined here: https://github.com/seL4/seL4/issues/214#issuecomment-1558628554. This is currently a WIP.


apps/sel4test-tests/src/tests/cache.c
```diff
@@ -245,11 +245,242 @@ static int test_page_directory_flush(env_t env)
...
+    vka_t *vka;
+    int err;
+
+    vka = &env->vka;
+
+    void *vaddr;
+    reservation_t reservation;
+
+    reservation = vspace_reserve_range(&env->vspace,
+                                       PAGE_SIZE_4K, seL4_AllRights, 1, &vaddr);
```
Indanz, line 282: I'd put this on one line, there are longer lines in the function.

apps/sel4test-tests/src/tests/cache.c
```diff
@@ -245,11 +245,242 @@ static int test_page_directory_flush(env_t env)
...
+ */
+static int test_cache_invalid(env_t env)
+{
+    seL4_CPtr frame;
+    uintptr_t vstart;
+    vka_t *vka;
+    int err;
+
+    vka = &env->vka;
+
```
Indanz, line 277: I'd combine this with the declaration, instead of putting it halfway other declarations.

apps/sel4test-tests/src/tests/cache.c
```diff
@@ -245,11 +245,242 @@ static int test_page_directory_flush(env_t env)
...
             config_set(CONFIG_HAVE_CACHE))
 DEFINE_TEST(CACHEFLUSH0003, "Test that cache maintenance can be done on large pages", test_large_page_flush_operation,
             config_set(CONFIG_HAVE_CACHE))
+DEFINE_TEST(CACHEFLUSH0004, "Test cache maintenance on invalid mappings", test_cache_invalid,
+            config_set(CONFIG_HAVE_CACHE))
+DEFINE_TEST(CACHEFLUSH0005, "Test cache maintenance on kernel-only mappings", test_cache_kernel_only,
+            config_set(CONFIG_HAVE_CACHE))
+DEFINE_TEST(CACHEFLUSH0006, "Test cache maintenance on read-write mappings", test_cache_read_write,
+            config_set(CONFIG_HAVE_CACHE))
+DEFINE_TEST(CACHEFLUSH0008, "Test cache maintenance on read-only mappings", test_cache_read_only,
```
Indanz, line 482: I think this should be CACHEFLUSH0007 and the last one 8 instead of 9.

But I think I would prefer these new tests at the end of the file instead of lumping it here, with the last test as the sole exception. That way test names stay the same in time, which can be useful in case people try to debug problems through git history.

Indanz (changes requested):
Other that the nitpicks everything seems good to go.

An alternative approach would be to have one test that remaps the same page instead, with one test function that takes two arguments: The expected result for invalidate and the expected result for all other operations. That would avoid a lot of duplicate code, but I'm not sure it's an overall improvement, as this version is very straightforward and easy to check.

apps/sel4test-tests/src/tests/cache.c
kent-mcleod: > But I think I would prefer these new tests at the end of the file instead of lumping it here, with the last test as the sole exception. That way test names stay the same in time, which can be useful in case people try to debug problems through git history.

I agree with preferring appending the test names instead of renaming where possible.

kent-mcleod (approved):
Looks good to me, thanks for doing this.

One observation is that the kernel doesn't track whether a mapping is cached or uncached. I wouldn't expect the architecture to have a problem with cache maintenance operations on mappings that are uncached, but maybe the seL4 design would wish to block cache maintenance operations on uncached mappings done through the vspace cap?

lsf37 (approved):
Also good from my side. I've switched on `hw-test`, because we should see some of these currently failing.

Indanz:
> One observation is that the kernel doesn't track whether a mapping is cached or uncached. I wouldn't expect the architecture to have a problem with cache maintenance operations on mappings that are uncached, but maybe the seL4 design would wish to block cache maintenance operations on uncached mappings done through the vspace cap?

If seL4 doesn't care and the hardware doesn't care, why bother?

kent-mcleod:
> If seL4 doesn't care and the hardware doesn't care, why bother?

It's not blocking now, but the point is that there could be an ambient authority situation where an uncached mapping implies an authority to manage caches for that mapped region. But you're right its probably a non-issue at this point because seL4 already doesn't care in other ways (ie there's no rights for restricting the ability to map something cached vs uncached).

apps/sel4test-tests/src/tests/cache.c
```diff
@@ -327,3 +327,235 @@ static int test_page_uncached_after_retype(env_t env)
...
+
+    /* Top-level page table operations */
+    err = seL4_ARCH_PageDirectory_Clean_Data(env->page_directory, vstart, vstart + PAGE_SIZE_4K);
+    test_error_eq(err, seL4_FailedLookup);
+    err = seL4_ARCH_PageDirectory_Invalidate_Data(env->page_directory, vstart, vstart + PAGE_SIZE_4K);
+    test_error_eq(err, seL4_FailedLookup);
+    err = seL4_ARCH_PageDirectory_CleanInvalidate_Data(env->page_directory, vstart, vstart + PAGE_SIZE_4K);
+    test_error_eq(err, seL4_FailedLookup);
+    err = seL4_ARCH_PageDirectory_Unify_Instruction(env->page_directory, vstart, vstart + PAGE_SIZE_4K);
+    test_error_eq(err, seL4_FailedLookup);
```
lsf37, line 381: VSpace cache operations on invalid mappings are silently ignored:

https://github.com/seL4/seL4/blob/1e6b69cdb63d83e4253340c1e13532681bf46cf1/src/arch/arm/64/kernel/vspace.c#L1338-L1346

Not sure I agree with the "astonishing" in the comment, but at least that was a conscious decision at some point and is part of the API, so I don't think we should force change there now.

lsf37:
@Indanz can you please check if you're now happy with Ivan's changes?

lsf37:
While we're expecting the tests to currently fail, it looks to me like some of the operations are returning a different kind of error, e.g. cap lookup fault instead of illegal operation.

Ivan-Velickovic:
> While we're expecting the tests to currently fail, it looks to me like some of the operations are returning a different kind of error, e.g. cap lookup fault instead of illegal operation.

yea I don't know what's going on. I'll have a look tomorrow.

apps/sel4test-tests/src/tests/cache.c
```diff
@@ -327,3 +327,229 @@ static int test_page_uncached_after_retype(env_t env)
...
+
+    reservation = vspace_reserve_range(&env->vspace, PAGE_SIZE_4K, seL4_AllRights, 1, &vaddr);
+    assert(reservation.res);
+
+    vstart = (uintptr_t)vaddr;
+    assert(IS_ALIGNED(vstart, seL4_PageBits));
+
+    /* Create a frame, but deliberately do not create a mapping for it. */
+    frame = vka_alloc_frame_leaky(vka, PAGE_BITS_4K);
+    test_assert(frame != seL4_CapNull);
```
Indanz, line 371: I'd move this into a separate function which with as parameter the rights to pass to `vspace_reserve_range` and call it in all the tests. That would save a lot of code duplication and make it easier to see the bits that change between the tests. At least I think it's all identical otherwise.

If not doing this, at least make all the code exactly the same without subtle differences for no reason, so it's easier to see it's identical.

apps/sel4test-tests/src/tests/cache.c
```diff
@@ -327,3 +327,229 @@ static int test_page_uncached_after_retype(env_t env)
...
+
+    /* Top-level page table operations for invalid mappings are silently ignored by
+     * the kernel so we do not test them here */
+
+    /* Page-level operations */
+    err = seL4_ARM_Page_Clean_Data(vstart, 0, PAGE_SIZE_4K);
```
Indanz, line 377: I don't understand how this can compile, `seL4_ARM_Page_Clean_Data` takes a page cap, start and end offsets.

lsf37: I think that's why we're getting cap faults -- `vstart` is not a cap index, so it looks up some random cap slot.

lsf37:
(rebased and tweaked under which configs we should expect an error)

lsf37:
Now passes on all runs for all boards. We still need to fix up the review comments, though.

Indanz (approved):

PR #97, opened 2023-07-24
Add test for AArch64 cache instructions

Ivan-Velickovic:
This patch adds a simple test to ensure that the instructions for cache maintenance from user-space succeed (when the kernel is configured to allow it).

It does not test that the cache operations produce the right side affects, just that they don't cause a fault/exception.

I did not add a test to check that a fault *does* occur if `CONFIG_AARCH64_USER_CACHE_ENABLE` is not enabled, since sel4test is only run with that config operation enabled.

### DIFF
--- a/apps/sel4test-tests/src/tests/cache.c
+++ b/apps/sel4test-tests/src/tests/cache.c
@@ -327,3 +327,52 @@ static int test_page_uncached_after_retype(env_t env)
 DEFINE_TEST(CACHEFLUSH0004, "Test that mapping a frame uncached doesn't see stale data after retype",
             test_page_uncached_after_retype,
             config_set(CONFIG_HAVE_CACHE))
+
+/*
+ * On AArch64, the kernel can allow the user to perform certain cache
+ * maintenance operations directly, without going into the kernel. The specific
+ * instructions are DC CVAU, DC CIVAC, DC CVAC, and IC IVAU.
+ *
+ * This test ensures that no exceptions arise from executing these instructions
+ * when the kernel is configured to allow them from user-space.
+ *
+ */
+static int test_user_cache_ops(env_t env)
+{
+    seL4_CPtr frame;
+    uintptr_t vstart;
+    vka_t *vka;
+    int err;
+
+    vka = &env->vka;
+
+    void *vaddr;
+    reservation_t reservation;
+
+    reservation = vspace_reserve_range(&env->vspace,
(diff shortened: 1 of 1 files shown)


apps/sel4test-tests/src/tests/cache.c
```diff
@@ -327,3 +327,52 @@ static int test_page_uncached_after_retype(env_t env)
...
+    frame = vka_alloc_frame_leaky(vka, PAGE_BITS_4K);
+    test_assert(frame != seL4_CapNull);
+
+    /* map in a cap with cacheability */
+    err = vspace_map_pages_at_vaddr(&env->vspace, &frame, NULL, vaddr, 1, seL4_PageBits, reservation);
+    test_error_eq(err, seL4_NoError);
+
+    /* Now that we have a page of memory in our virtual address space, we can
+     * test the cache instructions in its virtual address. */
+    asm volatile("dc cvau, %0" :: "r"(vaddr));
```
Ivan-Velickovic, line 369: I'm not super familiar with the sel4test infrastructure, is this the correct way to do this or should I be creating helper threads that execute the instructions and then calling `wait_for_helper`? If these instructions fail will it the whole sel4test run or will just this test fail?

kent-mcleod: Using a helper thread would be better IMO as it would allow you to catch the fault in the case where the user mode instructions were not enabled.  The way it is currently the test thread would fault and you can't tell whether the failure was due to the code-under-test or the test itself.  An inverse of the test would be to check that the instructions trigger an seL4_UserException fault if executed when the config option is not enabled.

Ivan-Velickovic: Okay cool I'll do it with a helper thread then. Not sure if you saw my comment in the PR description but is there much point in adding the inverse test since it will never be triggered by CI?

kent-mcleod: You could write the test in a way that it unconditionally runs and just checks that the thread result matches the configuration.  You could validate that the test passes for both configurations, and CI will test what ever configurations it tests, but because there's only one test implemented it's less likely to go stale for the other configuration but people can still run sel4test on their specific configuration even if the CI doesn't test it.

Ivan-Velickovic: Sure thing, I'll do add that then.

PR #99, opened 2023-07-31
Fix: NULL-terminate existing_frames[] array

heshamelmatary:
libsel4utils/reserve_initial_task_regions iterates over existing_frames[] and expects it to be null-terminated. Without this commit, the behaviour will rely on whatever the stack had before allocating existing_frames[] in the stack, and subsequent calls may fail, hang, or reserve incorrect frames; depending on the stack values.


axel-h (approved):
Thanks, nice catch. 

Looks good to me. Please rebase to latest master and address the gitlint checker remark.

kent-mcleod (approved):

PR #105, opened 2023-10-27
SCHED0021: decrease chances of a race condition

lsf37:
Decrease the chance that the monitor thread is preempted just when it has released the preemption threads, but before it has yielded itself. (If that happens, the preemption threads get two time slices.)

See also #42 -- the implementation here is basically @axel-h's suggestion from that issue.

We believe we might have seen an instance of this in one test run (of many) for PR seL4/seL4#968

### DIFF
--- a/apps/sel4test-tests/src/tests/scheduler.c
+++ b/apps/sel4test-tests/src/tests/scheduler.c
@@ -1598,11 +1598,17 @@ static int test_simple_preempt(struct env *env)
 
     /* Set a timeout for the test.
      * Each thread should be run for one tick */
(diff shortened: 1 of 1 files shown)


lsf37:
After this is merged, I might try enabling this test for simulation again in a separate PR to see if it got any better. Although I don't really think it was this specific race condition that was the reason for failure there.

Indanz (approved):

Indanz:
This should be enough to make it robust. If time_now() can't run within one time slice, there's something horribly wrong.

axel-h (approved):
Looks good to me

PR #111, opened 2024-01-15
Hardware debug API: Extend to aarch64 and add single stepping test

alwin-joshy:
seL4's hardware debug API is currently only implemented for ARMv7 and x86. I am currently implementing it for AARCH64 and as a part of this, have ported the software breakpoint test to AARCH64.

I have also implemented a new test to check that single stepping works on ARM. In doing so, I believe I have found a bug in the implementation which I will make another pull request for soon. This test is simpler than the existing x86 equivalent but I think it should be sufficient. If not, I can port the existing x86 test for single stepping to ARM.

Test with: https://github.com/seL4/seL4/pull/1162


apps/sel4test-tests/arch/arm/arch/debug.h
```diff
@@ -5,16 +5,36 @@
...
+        "label_two: \n\t" \
+        "nop \n\t" \
+        )
 
 /* Tell C about the symbols exported by the ASM above. */
-extern char sbreak;
+extern char sbreak, label_one, label_two;
 #define TEST_SOFTWARE_BREAK_EXPECTED_FAULT_LABEL sbreak
+#define TEST_SS_LABEL_ONE label_one
+#define TEST_SS_LABEL_TWO label_two
```
Indanz, line 37: Not sure why this indirection is done, but anyway, I would call it `TEST_SS_BREAKPOINT1` and 2.

apps/sel4test-tests/src/arch/arm/tests/breakpoints.c
```diff
@@ -0,0 +1,109 @@
...
+#include "../../../test.h"
+#include "../../../tests/breakpoints.h"
+
+static int faulter_main(seL4_Word arg0, seL4_Word arg1, seL4_Word arg2, seL4_Word arg3)
+{
+    TEST_SINGLE_STEP_ASM();
+    /* NOTHING BELOW SHOULD EVER BE REACHED */
+    int *ptr = NULL;
+    *ptr = 0xBEEF;
+    return 0;
```
Indanz, line 25: Maybe say something like: "Should not be reached, but if it does, it wakes up the handler thread".

apps/sel4test-tests/src/arch/arm/tests/breakpoints.c
```diff
@@ -0,0 +1,109 @@
...
+
+    /* Wait for the single step exception */
+    fault_data.vaddr = seL4_GetMR(seL4_DebugException_FaultIP);
+    fault_data.reason = seL4_GetMR(seL4_DebugException_ExceptionReason);
+    fault_data.vaddr2 = seL4_GetMR(seL4_DebugException_TriggerAddress);
+    fault_data.bp_num = seL4_GetMR(seL4_DebugException_BreakpointNumber);
+
+    if (label == seL4_Fault_DebugException &&
+        fault_data.reason == seL4_SingleStep &&
+        fault_data.vaddr == (seL4_Word) &TEST_SS_LABEL_TWO) {
```
Indanz, line 65: You could add this check for 'TEST_SS_LABEL_ONE' above too. If the test passes it doesn't matter, but if something went wrong with the initial breakpoint it would make debugging a lot easier.

apps/sel4test-tests/src/arch/arm/tests/breakpoints.c
```diff
@@ -0,0 +1,109 @@
...
+
+    /* Wait for the faulter to trigger an event. */
+    tag = api_wait(fault_ep_cspath.capPtr, &sender_badge);
+    ZF_LOGV("Handler got a fault on the ep.\n");
+    label = seL4_MessageInfo_get_label(tag);
+
+    /* Check that it is a breakpoint exception */
+    if (label != seL4_Fault_DebugException ||
+        seL4_GetMR(seL4_DebugException_ExceptionReason) != seL4_InstructionBreakpoint) {
+        ZF_LOGV("Received exception other than instruction breakpoint exception.\n");
```
Indanz, line 42: Maybe explicitly test for NULL and say that single stepping doesn't work.

Indanz (approved):
Looks good, thanks.

https://github.com/seL4/seL4/pull/1162 is about ARMv7, not ARMv8. Wrong pull?

Indanz:
With 1162 in mind, I think it would be good to test whether code actually gets executed instead of only checking if the instruction pointer advances, so instead of nops just add a value to a register and check if it's correct at the second label.

alwin-joshy:
> With 1162 in mind, I think it would be good to test whether code actually gets executed instead of only checking if the instruction pointer advances, so instead of nops just add a value to a register and check if it's correct at the second label.

Good point. In that case, I'll port the equivalent x86 test, which seems to be more robust.

Indanz:
> > With 1162 in mind, I think it would be good to test whether code actually gets executed instead of only checking if the instruction pointer advances, so instead of nops just add a value to a register and check if it's correct at the second label.
> 
> Good point. In that case, I'll port the equivalent x86 test, which seems to be more robust.

 At first glance I don't really see anything x86 specific in there, perhaps make that test cross-platform with some arch specific helper functions or defines to paper over the differences? That definitely seems to be the current intention. You'd need to move the test to some common place and make sure it still gets executed on x86.

alwin-joshy:
> > > With 1162 in mind, I think it would be good to test whether code actually gets executed instead of only checking if the instruction pointer advances, so instead of nops just add a value to a register and check if it's correct at the second label.
> > 
> > 
> > Good point. In that case, I'll port the equivalent x86 test, which seems to be more robust.
> 
> At first glance I don't really see anything x86 specific in there, perhaps make that test cross-platform with some arch specific helper functions or defines to paper over the differences? That definitely seems to be the current intention. You'd need to move the test to some common place and make sure it still gets executed on x86.

Yeah, should be simple. I do recall seeing a comment somewhere in the code about something being missing or not ready in ARM, but I'll see if I run into any issues. Didn't spot anything that immediately stood out.

alwin-joshy:
@Indanz I moved the x86 breakpoint test into the arch-independent breakpoints file and it just worked.

lsf37:
It looks like the `hw-test` trigger isn't working properly on this repo. I'll investigate..

Indanz:
The Sabre clang fails with:

```
   All is well in the universe
  
  
  
  console_run returned 0
  
  0.05user 0.02system 2:48.06elapsed 0%CPU (0avgtext+0avgdata 7936maxresident)k
  0inputs+336outputs (0major+8920minor)pagefaults 0swaps
  +++ python3 ../projects/seL4_libs/libsel4test/tools/extract_results.py -q results.xml parsed_results.xml
  No log to check boot failure on.
  +++ mq.sh sem -signal sabre2 -k seL4/sel4test-seL4Test HW-7619487774-hw-run-5
  Error parsing parsed_results.xml
  -----------[ end test SABRE_debug_MCS_clang_32 ]-----------
SABRE_debug_MCS_clang_32 FAILED
```

Indanz:
I verified that it still runs and passes on x86 and some 32 bit ARMs, it doesn't yet run on Aarch64 though. I think we should remove the lines around https://github.com/seL4/seL4/blob/cc3205ea486f6ce3da3a64d8a99e166f047b8419/src/arch/arm/config.cmake#L226.

Indanz:
So do we want to remove `KernelHardwareDebugAPIUnsupported` from aarch64 before merging this, as the aarch64 code isn't tested yet, or merge this first and remove the define later?

@alwin-joshy, did you test this on aarch64 yet?

alwin-joshy:
> I verified that it still runs and passes on x86 and some 32 bit ARMs, it doesn't yet run on Aarch64 though. I think we should remove the lines around https://github.com/seL4/seL4/blob/cc3205ea486f6ce3da3a64d8a99e166f047b8419/src/arch/arm/config.cmake#L226.

Yep, that will be necessary for getting the test to run on AARCH64 eventually but there are other substantial changes required for this. I've been working on this and am close to done (https://github.com/alwin-joshy/seL4/pull/7) but there are still a few problems related to ARMv8 32-bit and ARM hyp mode that I want to iron out before submitting a proper PR.

alwin-joshy:
> So do we want to remove `KernelHardwareDebugAPIUnsupported` from aarch64 before merging this, as the aarch64 code isn't tested yet, or merge this first and remove the define later?
> 
> @alwin-joshy, did you test this on aarch64 yet?

I think to merge this first and remove the KernelHardwareDebugAPIUnsupported in the PR when I have the hardware debug API fully ported to aarch64

apps/sel4test-tests/arch/arm/arch/debug.h
```diff
@@ -5,12 +5,21 @@
...
 #pragma once
 
+#ifdef CONFIG_ARCH_AARCH32
 #define TEST_SOFTWARE_BREAK_ASM() \
     asm volatile( \
         ".global sbreak, post_sbreak\n\t" \
         ".type post_sbreak, function\n\t" \
         "sbreak:\n\t" \
         "bkpt\n\t")
+#else
```
Indanz, line 15: Perhaps make it an explicit check for AARCH64, to make it clearer what's happening.

```suggestion
#elif CONFIG_ARCH_AARCH64
```

If you tested this assembly locally I'm happy to merge this PR.

alwin-joshy: Yep, the test passes with the changes I've made in https://github.com/alwin-joshy/seL4/pull/7

PR #116, opened 2024-02-20
SMC0001: Remove test of clobbered registers

kent-mcleod:
ATF implementations don't clobber x2 and x3 registers in response to ARM_STD_SVC_VERSION calls.


kent-mcleod:
https://github.com/ARM-software/arm-trusted-firmware/blob/514d022fdae9305d4381bbf7ad19db878cfaa8eb/include/arch/aarch64/smccc_helpers.h#L40-L42

lsf37 (approved):

PR #117, opened 2024-03-31
Allow setting NANOPB_SRC_ROOT_FOLDER 

axel-h:
- in CMake, `INTERNAL` has the same effect as `FORCE`, so passing any `-DNANOPB_SRC_ROOT_FOLDER=...` to CMake on the command line will be overwritten and there is no way to customize the path then.
- the second commit removes the trailing slash of the variable, which seem more a cosmetic thing. Having a "//" on some paths in the variables that are created later seems not to cause any practical issues.

### DIFF
--- a/settings.cmake
+++ b/settings.cmake
@@ -17,7 +17,7 @@ list(
         ${project_modules}
 )
 
-set(NANOPB_SRC_ROOT_FOLDER "${project_dir}/tools/nanopb" CACHE INTERNAL "")
+set(NANOPB_SRC_ROOT_FOLDER "${project_dir}/tools/nanopb" CACHE STRING "NanoPB Folder location")
 set(OPENSBI_PATH "${project_dir}/tools/opensbi" CACHE STRING "OpenSBI Folder location")
 
 set(SEL4_CONFIG_DEFAULT_ADVANCED ON)
--- a/settings.cmake
+++ b/settings.cmake
@@ -6,7 +6,7 @@
 
 cmake_minimum_required(VERSION 3.7.2)
 
-set(project_dir "${CMAKE_CURRENT_LIST_DIR}/../../")
+set(project_dir "${CMAKE_CURRENT_LIST_DIR}/../..")
 file(GLOB project_modules ${project_dir}/projects/*)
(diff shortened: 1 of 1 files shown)


Indanz (approved):
Out of curiosity, why are you using a different NanoPB and why can't you put it in tools/nanopb instead? Or symlink tools/nanopb to wherever your nanopb is?

axel-h:
I am using the same nanoPB version, but just a different directory layout. Symlinking is not really an option, as this would be the only thing that needs the project workspace and tools folder then.

I did not get things to work with just adding nanoPB to the `CMAKE_MODULE_PATH`. Making this work could be another mire generic way, but it may need more changes.

Indanz:
The directory layout seems quite hard-coded, how do you manage the other tools paths? Is it worth trying to re-use settings.cmake? It doesn't seem to be doing that much useful work.

axel-h:
Providing `-DCMAKE_MODULE_PATH=...` works well enough for all other modules.

axel-h:
Concerning your question about using `settings.cmake` or not, that is where https://github.com/seL4/sel4test/pull/40 is aiming for to allow full customization. But as stated there, `settings.cmake` contains some useful parts, so using it has some benefits.

axel-h:
Since `sel4bech` basically uses the same `settings.cmake`, but does not need the patch, I have been digging a bit more into this. It seems `sel4test`is running in to the issues described at https://stackoverflow.com/questions/75619926/does-cmake-set-internal-apply-force because CMake runs `settings.cmake` twice here (at least when targeting QEMU, the logs show this). The second runs finally makes `INTERNAL` act in `FORCE`-mode and overwrite the command line setting. In `sel4bench` it runs only once, so the command line value is used.
I still think we should not use `INTERNAL` here (and also apply this patch in `sel4bench`). There is a message we print in https://github.com/seL4/seL4_tools/blob/8c660fd6d8ba60aa4c76d5c9586e3892502d8e0e/cmake-tool/helpers/nanopb.cmake#L22 that also implies this values is supposed to be be something one can set on the command line.

PR #122, opened 2024-05-28
remove explicit newline in ZF_LOGx()

axel-h:
- remove explicit newline in `ZF_LOGx()`, the functions already prints a newline at the end.
- SCHED0003: log debug information instead of empty lines


lsf37 (approved):
Looks good from my side. I've been wondering why we are adding all those empty lines. Did the ZF_LOGV implementation change at some point?

axel-h:
Some might have been `printf()` once. Or just sloppyness.

PR #125, opened 2024-07-24
add test for round robin threads

lsf37:
- add test SCHED0023 to check that round robin threads are indeed scheduled round robin.
- print thread id in periodic thread tests
 
The test relies on relative timing of a busy wait loop. The loop has to take at least 1 ms so that the round-robin test thread is preempted at least once before it prints+increments again.


apps/sel4test-tests/src/tests/scheduler.c
```diff
@@ -1695,3 +1695,72 @@ static int test_changing_affinity(struct env *env)
...
+
+    /* assumes counters are initialised already */
+    while (counters[id] < SCHED0023_NUM_ROUNDS) {
+        counters[id]++;
+        /* thread "id" should run after thread "id-1", so we should be at the same
+           count after incrementing if we are scheduled round robin. */
+        if (0 < id) {
+            test_eq(counters[id], counters[id - 1]);
+        }
+        printf("Tick (%d)\n", id);
```
axel-h, line 1722: Maybe add a explicit note here that the blocking printing is not interfering with the test.

apps/sel4test-tests/src/tests/scheduler.c
```diff
@@ -1695,3 +1695,72 @@ static int test_changing_affinity(struct env *env)
...
+        /* busy wait so we can be preempted at least once before we increment
+           the counter again */
+        delay_loop(5000000);
+    }
+
+    test_eq(counters[id], (unsigned long) SCHED0023_NUM_ROUNDS);
```
axel-h, line 1728: Maybe add a note there why this would ever fail. Is there any other cause besides memory corruption?

apps/sel4test-tests/src/tests/scheduler.c
```diff
@@ -1695,3 +1695,72 @@ static int test_changing_affinity(struct env *env)
...
+        __asm__ __volatile__ ("");
+    }
+}
+
+#define SCHED0023_NUM_ROUNDS 5
+
+/* used in test_round_robin below */
+static void round_robin_thread(int id, volatile unsigned long *counters)
+{
+    delay_loop(id * 100);
```
axel-h, line 1712: Why do we need this delay actually, is it to ensure `counters` is initialized and visible?

apps/sel4test-tests/src/tests/scheduler.c
```diff
@@ -1695,3 +1695,72 @@ static int test_changing_affinity(struct env *env)
...
+        counters[id]++;
+        /* thread "id" should run after thread "id-1", so we should be at the same
+           count after incrementing if we are scheduled round robin. */
+        if (0 < id) {
+            test_eq(counters[id], counters[id - 1]);
+        }
+        printf("Tick (%d)\n", id);
+        /* busy wait so we can be preempted at least once before we increment
+           the counter again */
+        delay_loop(5000000);
```
axel-h, line 1725: Maybe add a note here why this magic value seem quite save based on the assumption about time slice and core frequency/IPS.

apps/sel4test-tests/src/tests/scheduler.c
lsf37: Just making sure every thread is only writing to its own counter.

apps/sel4test-tests/src/tests/scheduler.c
lsf37: It’s to make sure that the writes to the counters are in order. When the threads are first started, this is not necessary, but when each thread gets scheduled the second time, we’d be relying on that each of them got the exact same number of cycles to make progress. If thread 0 got just 10 cycles less, the second counter increment in thread 0 might happen after the second counter increment in thread 1. I don’t think the timer interrupts are that accurate for all platforms, so I’m adding some safety.

apps/sel4test-tests/src/tests/scheduler.c
lsf37: It looks like it isn’t safe yet on all machines, esp when printing is off. I might have to actually try to compute a value here instead. Just wanted to avoid making any syscalls in these threads that could affect scheduling.

apps/sel4test-tests/src/tests/scheduler.c
axel-h: Good point. Please add this in a comment in the code.

apps/sel4test-tests/src/tests/scheduler.c
Indanz: I think it will interfere with the test, as it messes up all timing and hence scheduling.

Indanz (commented):
The whole test seems incredibly fragile.

The test should be inherently stable, so the busy looping at the start should be unnecessary. But if you do loop, at least add a nop or something, in case empty asm strings are optimised away.

To start the test I wouldn't lower the priority of the main thread, but instead let the main thread block till all tests are done. But if you do, please increase it again at the end of the test. I'd let the main thread start and block on a timer so the whole test has a fixed runtime (or use MCS with period = test period and budget -1 us + yield as a timer for the main thread).

Why not increment and check the counters in a busy loop? Check that the counter values are always lower than the previous task's one.

To avoid the problem of small jitter making tasks overtaking each other, you can give them progressively slightly shorter periods, but with differences small enough that a non-round-robin schedule is detected. You probably want to increase the period for this, as that reduces the jitter and scheduling overhead relatively.

PR #128, opened 2024-11-24
Speedup Tests

Indanz:
This more than halves the time to run all tests.

### DIFF
--- a/apps/sel4test-driver/src/tests/timer.c
+++ b/apps/sel4test-driver/src/tests/timer.c
@@ -39,7 +39,7 @@ int test_timer(driver_env_t env)
     int error = tm_alloc_id_at(&env->tm, TIMER_ID);
     test_assert_fatal(!error);
 
-    error = tm_register_cb(&env->tm, TIMEOUT_PERIODIC, 1 * NS_IN_S, 0, TIMER_ID,
+    error = tm_register_cb(&env->tm, TIMEOUT_PERIODIC, 100 * NS_IN_MS, 0, TIMER_ID,
                            test_callback, (uintptr_t) &test_data);
     test_assert_fatal(!error);
 
--- a/apps/sel4test-tests/src/tests/faults.c
+++ b/apps/sel4test-tests/src/tests/faults.c
@@ -694,74 +694,73 @@ static int test_fault(env_t env, int fault_type, bool inter_as)
     vka_object_t reply;
     error = vka_alloc_reply(&env->vka, &reply);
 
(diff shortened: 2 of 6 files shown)


Indanz:
All checks passed, let's try hw tests now.

Indanz:
Only test failing is `MULTICORE0005` on gcc TX2 and I didn't change that one.

Indanz:
The main time is spend in CI overhead and board bring up, so running the tests in CI still takes forever. :-(

kent-mcleod:
> The main time is spend in CI overhead and board bring up, so running the tests in CI still takes forever. :-(

In some cases the board bring-up times can usually be optimized because each board is run with a custom run script that usually introduces the delays.

But some boards take a long time from power on to the u-boot prompt. This could also be optimized but someone would need to rebuild the boards boot firmware.

I think the CI overhead is still what dominates though.

kent-mcleod (approved):

Indanz:
> But some boards take a long time from power on to the u-boot prompt. This could also be optimized but someone would need to rebuild the boards boot firmware.

Switching from tftp boot to USB boot via UUU for NXP based SoCs would probably save a fair bit. Newer SoCs also seem to have built-in board control and can e.g. be power cycled via USB with https://github.com/nxp-imx/bcu.

> I think the CI overhead is still what dominates though.

I think there are some suboptimal things going on there. For instance, the last CI had to wait an hour for a lock to become free. The lock was held by the next CI run. That seems fairly bizarre to me. On one hand CI runs release the lock in between all tests, on the other hand it can block a board for an hour?

Thanks for the review!

Ivan-Velickovic:
> Switching from tftp boot to USB boot via UUU for NXP based SoCs would probably save a fair bit. Newer SoCs also seem to have built-in board control and can e.g. be power cycled via USB with https://github.com/nxp-imx/bcu.

I've noticed that the ethernet drivers in U-Boot are generally quite poor so for a lot of boards it can take up to a half a minute or so to just load an image (although for sel4test the images are fairly small so maybe it doesn't matter).

Given that the USB drivers would also be polling-based would it actually be any better?

Indanz:
> Given that the USB drivers would also be polling-based would it actually be any better?

I don't know, I set the board in serial boot mode and use UUU to upload SPL+Uboot,
then start Uboot in Fastboot mode and run Uboot commands from UUU, including
loading the image and booting it. All that takes 4 seconds from power-on to the first
Elfloader print.

I can share the UUU script I'm using if anyone is interested.

PR #133, opened 2025-02-12
vspace004: allocate a cap for the final pool

midnightveil:
This *coincidentally* fails with ret == seL4_RevokeFirst(9) if the number of pools to allocate before failing is less than the actual amount, because we don't allocate a capability for the last pool.

In essence, that could be happening even if we had the correct number of pools, but doesn't because of how the kernel (presumably) checks that the cap is free after checking the ASID pools are free.


Indanz (approved):
At this point it might be cleaner to bring the last check within the loop and check `i` to determine the expected `ret`, but this is fine too.

Indanz:
Yes, it looks very suspicious, I thought we had most tests stable now with the recent SMP fixes, not sure what's going on here. Why do only some configs fail? Did a recent commit break something?

lsf37 (approved):
The current hw test failures are unrelated, so fine to merge from my side.

PR #135, opened 2025-02-27
arm,cmake: Set compiler expected tls settings

kent-mcleod:
Add the right TLS variable settings for the gnu-elf-abi used by our compilers.


Indanz (commented):
Does sel4test itself really care one way or the other? Shouldn't it follow the seL4 defaults?

PR #145, opened 2025-07-23
Fix test failures when -fshort-enums is used

midnightveil:
When compiled with arm-none-eabi-gcc (13.3.0, and others), the default ABI has -fshort-enums on by default. This seem to also be the case for x86 with gcc (GCC) 15.1.1 20250521 (Red Hat 15.1.1-2).

Due to C's arcane integer promotion rules, the enum members such as SUCCESS or seL4_NoError are of type "int" (generally signed 32 bit), whereas the size of seL4_Error or test_result_t is 8 bits (a char).

For reference: https://godbolt.org/z/zoKzhqxc7

This produces test output like:

    Error: res (size 1) != SUCCESS (size 4), use of test_eq incorrect

Test with: https://github.com/seL4/seL4_libs/pull/105


apps/sel4test-tests/src/tests/faults.c
```diff
@@ -1080,7 +1080,7 @@ static int test_vm_enter_non_vm(env_t env)
     err = api_tcb_set_space(get_helper_tcb(&helper), fault_ep, env->cspace_root,
                             api_make_guard_skip_word(guard),
                             env->page_directory, seL4_NilData);
-    test_eq(err, 0);
+    test_eq(err, (seL4_Error)seL4_NoError);
```
lsf37, line 1083: The other occurrences of `seL4_NoError` in this repo are just `test_eq(err, seL4_NoError)`. The type of `seL4_NoError` is already the enum, so it looks to me like we should not need an explicit cast here. Or is there something special going on in this one?

Indanz (commented):
I acknowledge the problem, but I disagree with the solution: Surely the solution is to remove the sizeof check in the `test_eq` macro? What does that check add?

midnightveil:
The other alternative is we just pass through `-fno-short-enums` through the build system?

midnightveil:
Hm.

```
/nix/store/fj2jmvq3azd8lqqa6jw7qkbjnq33rdn0-arm-none-eabi-binutils-2.43.1/bin/arm-none-eabi-ld: warning: apps/sel4test-driver/CMakeFiles/sel4test-driver.dir/src/arch/arm/arch.c.obj uses 32-bit enums yet the output is to use variable-size enums; use of enum values across objects may fail
/nix/store/fj2jmvq3azd8lqqa6jw7qkbjnq33rdn0-arm-none-eabi-binutils-2.43.1/bin/arm-none-eabi-ld: warning: apps/sel4test-driver/CMakeFiles/sel4test-driver.dir/src/main.c.obj uses 32-bit enums yet the output is to use variable-size enums; use of enum values across objects may fail
/nix/store/fj2jmvq3azd8lqqa6jw7qkbjnq33rdn0-arm-none-eabi-binutils-2.43.1/bin/arm-none-eabi-ld: warning: apps/sel4test-driver/CMakeFiles/sel4test-driver.dir/src/tests/interrupt.c.obj uses 32-bit enums yet the output is to use variable-size enums; use of enum values across objects may fail
/nix/store/fj2jmvq3azd8lqqa6jw7qkbjnq33rdn0-arm-none-eabi-binutils-2.43.1/bin/arm-none-eabi-ld: warning: apps/sel4test-driver/CMakeFiles/sel4test-driver.dir/src/tests/smmu.c.obj uses 32-bit enums yet the output is to use variable-size enums; use of enum values across objects may fail
/nix/store/fj2jmvq3azd8lqqa6jw7qkbjnq33rdn0-arm-none-eabi-binutils-2.43.1/bin/arm-none-eabi-ld: warning: apps/sel4test-driver/CMakeFiles/sel4test-driver.dir/src/tests/syscall.c.obj uses 32-bit enums yet the output is to use variable-size enums; use of enum values across objects may fail
/nix/store/fj2jmvq3azd8lqqa6jw7qkbjnq33rdn0-arm-none-eabi-binutils-2.43.1/bin/arm-none-eabi-ld: warning: apps/sel4test-driver/CMakeFiles/sel4test-driver.dir/src/tests/timer.c.obj uses 32-bit enums yet the output is to use variable-size enums; use of enum values across objects may fail
/nix/store/fj2jmvq3azd8lqqa6jw7qkbjnq33rdn0-arm-none-eabi-binutils-2.43.1/bin/arm-none-eabi-ld: warning: apps/sel4test-driver/CMakeFiles/sel4test-driver.dir/src/testtypes.c.obj uses 32-bit enums yet the output is to use variable-size enums; use of enum values across objects may fail
/nix/store/fj2jmvq3azd8lqqa6jw7qkbjnq33rdn0-arm-none-eabi-binutils-2.43.1/bin/arm-none-eabi-ld: warning: apps/sel4test-driver/CMakeFiles/sel4test-driver.dir/src/timer.c.obj uses 32-bit enums yet the output is to use variable-size enums; use of enum values across objects may fail
```

This happens for every object linked in, regardless of if I add `-fno-short-enums` to cmake-tool or to sel4test (the latter makes it only complain about sel4test objects), or even if I add `-mabi=aapcs-linux`.

```
 /nix/store/yn7f2hdccdzclaz43zz7pi3akrn6v2gw-arm-none-eabi-gcc-13.3.0/bin/arm-none-eabi-gcc --sysroot=/home/julia/ts/sel4test-manifest-5/build  -march=armv7ve -marm  -D__KERNEL_32__ -g -D__KERNEL_32__   -static -nostdlib -z max-page-size=0x1000    -Wl,-u_sel4_start -Wl,-e_sel4_start
  -Wl,-T /home/julia/ts/sel4test-manifest-5/tools/seL4/cmake-tool/helpers/tls_rootserver.lds   -Wl,-umuslcsys_init_muslc /home/julia/ts/sel4test-manifest-5/build/lib/crt0.o /home/julia/ts/sel4test-manifest-5/build/lib/crti.o /nix/store/yn7f2hdccdzclaz43zz7pi3akrn6v2gw-arm-none-eabi-gcc-13.3.0/lib/gcc/arm-none-eabi/13.3.0/crtbegin.o apps/sel4test-driver/archive.o apps/sel4test-driver/CMakeFiles/sel4test-driver.dir/src/arch/arm/arch.c.obj apps/sel4test-driver/CMakeFiles/sel4test-driver.dir/src/main.c.obj apps/sel4test-driver/CMakeFiles/sel4test-driver.dir/src/tests/interrupt.c.obj apps/sel4test-driver/CMakeFiles/sel4test-driver.dir/src/tests/smmu.c.obj apps/sel4test-driver/CMakeFiles/sel4test-driver.dir/src/tests/syscall.c.obj apps/sel4test-driver/CMakeFiles/sel4test-driver.dir/src/tests/timer.c.obj apps/sel4test-driver/CMakeFiles/sel4test-driver.dir/src/testtypes.c.obj apps/sel4test-driver/CMakeFiles/sel4test-driver.dir/src/timer.c.obj -Wl,--start-group         -lgcc  libsel4/libsel4.a  apps/sel4test-driver/sel4runtime/libsel4runtime.a  apps/sel4test-driver/seL4_libs/libsel4allocman/libsel4allocman.a  apps/sel4test-driver/seL4_libs/libsel4vka/libsel4vka.a  apps/sel4test-driver/seL4_libs/libsel4utils/libsel4utils.a  apps/sel4test-driver/sel4_projects_libs/libsel4rpc/libsel4rpc.a  apps/sel4test-driver/seL4_libs/libsel4test/libsel4test.a  apps/sel4test-driver/seL4_libs/libsel4platsupport/libsel4platsupport.a  apps/sel4test-driver/seL4_libs/libsel4muslcsys/libsel4muslcsys.a  apps/sel4test-driver/libsel4testsupport/libsel4testsupport.a  -Wl,-u -Wl,__vsyscall_ptr  apps/sel4test-driver/seL4_libs/libsel4test/libsel4test.a  apps/sel4test-driver/sel4_projects_libs/libsel4rpc/libsel4rpc.a  apps/sel4test-driver/sel4_projects_libs/libsel4nanopb/libsel4nanopb.a  apps/sel4test-driver/sel4_projects_libs/libsel4nanopb/libnanopb.a  apps/sel4test-driver/seL4_libs/libsel4sync/libsel4sync.a  apps/sel4test-driver/seL4_libs/libsel4serialserver/libsel4serialserver.a  apps/sel4test-driver/seL4_libs/libsel4utils/libsel4utils.a  apps/sel4test-driver/util_libs/libelf/libelf.a  apps/sel4test-driver/util_libs/libcpio/libcpio.a  apps/sel4test-driver/seL4_libs/libsel4platsupport/libsel4platsupport.a  apps/sel4test-driver/sel4runtime/libsel4runtime.a  apps/sel4test-driver/util_libs/libplatsupport/libplatsupport.a  apps/sel4test-driver/util_libs/libfdt/libfdt.a  -Wl,--undefined=arm_gic_ptr,--undefined=tegra_ictlr_ptr,--undefined=arm_gicv3_ptr,--undefined=fsl_avic_ptr,--undefined=ti_omap3_ptr  apps/sel4test-driver/seL4_libs/libsel4simple-default/libsel4simple-default.a  apps/sel4test-driver/seL4_libs/libsel4vspace/libsel4vspace.a  apps/sel4test-driver/seL4_libs/libsel4simple/libsel4simple.a  apps/sel4test-driver/seL4_libs/libsel4vka/libsel4vka.a  apps/sel4test-driver/seL4_libs/libsel4debug/libsel4debug.a  libsel4/libsel4.a  apps/sel4test-driver/util_libs/libutils/libutils.a  apps/sel4test-driver/musllibc/build-temp/stage/lib/libc.a -Wl,--end-group /nix/store/yn7f2hdccdzclaz43zz7pi3akrn6v2gw-arm-none-eabi-gcc-13.3.0/lib/gcc/arm-none-eabi/13.3.0/crtend.o /home/julia/ts/sel4test-manifest-5/build/lib/crtn.o -o apps/sel4test-driver/sel4test-driver -Wl,--verbose
```

There's a way to hide this (`--no-enum-size-warning`) but I don't know if that's necessarily correct.  From the looks of things we're asking for `-lgcc` so libgcc would likely be a source of complaints: `/nix/store/yn7f2hdccdzclaz43zz7pi3akrn6v2gw-arm-none-eabi-gcc-13.3.0/lib/gcc/arm-none-eabi/13.3.0/libgcc.a`. It looks like BFD picks up the output attributes from the first input file it's given, which happens to be `crt0.o`.

Not sure if I like that approach now.

lsf37:
I've also hit this now (trying to build without docker for a change), and I don't see a way to remove the warning without just suppressing them, which does not sound like a good idea.

Overall, I'd actually be happy with the casts as in the original PR. From everything we've been discussing, they are the lowest impact change. Trying to re-architect what `test_eq` does looks like a deeper rabbit hole to me, and it's not like the casts are doing anything wrong. If at all, they are a bit of documentation on what is going on. Maybe we should add a comment that explains why we have them in a way that can be discovered via `grep` if someone else is hitting the same problem elsewhere in sel4test.

lsf37 (approved):

Indanz:
My original objections still stand:

> We have a wrong assumption in the check macro's (because they're part of the same code base that defined SUCCESS), which extends to all comparisons with enums, so to me the simple, low risk solution seems to remove those checks. That won't break any existing code and I don't see how it can cause future code to break.
>
> If we keep that arbitrary size check, all users will stumble into this subtle enum size thing on some platforms with some toolchains. The static assert won't catch problems before they get into the code, only afterwards. So all it does is break the tests now and then for users for no good reason, like what happened to you now. The static assert just makes it easier to debug, it doesn't avoid the problem.
> 
> Removing the size checks does avoid and solve the problem, now and in the future.

The root cause is the wrong assumption in the check macros, anything else is one work-around or another for that problem.

lsf37:
> > Removing the size checks does avoid and solve the problem, now and in the future.
> 
> The root cause is the wrong assumption in the check macros, anything else is one work-around or another for that problem.

So you'd remove the size assertion in here? https://github.com/seL4/seL4_libs/blob/4ce71fca2adbf1c2add7eabf5670633a70177a2f/libsel4test/include/sel4test/test.h#L239-L266

Wouldn't that make it more likely that something even more obscure is going wrong in all these `TYPES_COMPATIBLE` dispatch checks?

midnightveil:
I think we'd need to add two pairs of TYPES_COMPATIBLE checks then: one for a and one for b. That or I guess we really should be checking TYPES_COMPATIBLE(a, b) which I don't know why we didn't do originally instead of the size. [ It's not, strictly speaking, transitive, because whilst it might be the case that "enum X" ~ "int" and "enum Y" ~ "int", it is not the case that "enum X" ~ "enum Y" ]. Otherwise we'd run into the issue now I think where we see the short enum on the left, use the short enum printf specifier, but then also use it for the int on the right, since dispatch goes off the type of A. I think this is fine because varargs puts everything on the stack as an int, but I don't know if that's technically UB to do that or not.

We really want 2 things:

1. the typeof(a), then the typeof(b), for the right printf format string. This can be solved by making test_op do the printf formatting for a & b separately. This can likely be improved by using C11 _Generic to go from type -> "%llu" etc, and so that twice as arguments. (I say _Generic of the chain of if types compatible because then we can compile time error if we don't know how to print a particular type).

2. `typeof(a)` and `typeof(b)` are either both signed or unsigned, for inequality. (see: a = "int" and b = "unsigned int" has edge cases). Checking types compatible of a & b would be good enough except it's also too restrictive, as it misses the short-enums enum compatible with int issue. However, I don't think anywhere in the current code really cares about this, and if it did it's broken currently anyway since we were previously assuming that if it's the same size it's fine; and most of the time we use equality anyway in these test suites.

In most other cases e.g. equality of numbers the default C promotion to int, or to the larger of a and b for binary operators should be fine.

tldr: I think it's mostly about the printing. This would be a rework of the test_op macro, but it wouldn't break users, I don't think.

Indanz:
> So you'd remove the size assertion in here? https://github.com/seL4/seL4_libs/blob/4ce71fca2adbf1c2add7eabf5670633a70177a2f/libsel4test/include/sel4test/test.h#L239-L266

Yes.

> Wouldn't that make it more likely that something even more obscure is going wrong in all these `TYPES_COMPATIBLE` dispatch checks?

Maybe, but very unlikely. Worst case the top bits of `b` get discarded in the value print, making debugging slightly harder.

I find all that macro code overly complex for no gain. It does all the type checking, and then it still needs to cast things.

The sad thing is, because of those macros, you get punished for using the right enum type instead of `int`. The reason this problem is fixed by just changing those four lines is because that are the only places using `seL4_Error` and `test_result_t` instead of `int`.

If you don't want to remove the size check, changing these locations to also use `int`  would be more consistent with all the other code.

lsf37:
Ok, fine, let's fix the macro. In our own code, these macros are used exclusively in `sel4test`, `libsel4test` and `libsel4serialserver` (also for testing), we probably don't need to be extra super careful about compatibility.

@midnightveil, your plan sounds good to me. Do you have time to give this a go?

midnightveil:
> @midnightveil, your plan sounds good to me. Do you have time to give this a go?

Yeah, I can give this a shot today.

(I think this PR is the easier fix, it's not like this code needs to be perfect, but I can fix it properly too)

midnightveil:
Closing this in favour of https://github.com/seL4/seL4_libs/pull/111.

PR #148, opened 2025-10-20
Tests for overmapping, and fixes for aarch32 PT tests

Kswin01:
This is a rebased version of the work done in PR #88

Test with https://github.com/seL4/seL4/pull/1526

### DIFF
--- a/apps/sel4test-tests/src/tests/pagetables.c
+++ b/apps/sel4test-tests/src/tests/pagetables.c
@@ -657,9 +657,9 @@ static int test_overmap_small(env_t env)
     cspacepath_t path;
     int error;
 
-    seL4_CPtr pgd = vka_alloc_object_leaky(&env->vka, seL4_ARM_PageGlobalDirectoryObject, 0);
-    seL4_CPtr pud = vka_alloc_object_leaky(&env->vka, seL4_ARM_PageUpperDirectoryObject, 0);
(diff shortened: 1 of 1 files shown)


apps/sel4test-tests/src/tests/pagetables.c
```diff
@@ -508,3 +511,474 @@ test_pagetable_arm(env_t env)
...
+DEFINE_TEST(PT0014, "Test that overmapping super sections is not permitted", test_overmap_supersection, true)
+
+#elif defined(CONFIG_ARCH_AARCH64)
+static int test_overmap_small(env_t env)
+{
+    seL4_Word map_addr = 0x10000000;
+    cspacepath_t path;
+    int error;
+
+    seL4_CPtr pgd = vka_alloc_object_leaky(&env->vka, seL4_ARM_PageGlobalDirectoryObject, 0);
```
Indanz: Since Kent's commit to unify the page table objects, this should be `seL4_ARM_VSpaceObject` and the others just `seL4_ARM_PageTable`.

However, backward compatible defines are in `libsel4/sel4_arch_include/aarch64/sel4/sel4_arch/deprecated.h`. (But they're missing from `libsel4/sel4_arch_include/aarch32/sel4/sel4_arch/deprecated.h`, not sure if that's intentional.)

PR #151, opened 2026-03-02
Update domain tests for RFC-20

Indanz:
Test with https://github.com/seL4/seL4/pull/1511.


lsf37 (approved):

Indanz:
Maybe trivial, but I think you meant to say CMake style instead of camke style. Too late to change though. :-)

lsf37:
Indeed I did. Sorry about that.

apps/sel4test-tests/src/tests/domains.cxx
```diff
@@ -31,6 +148,8 @@ own_domain_success(struct env *env)
 
     error = seL4_DomainSet_Set(env->domain, CONFIG_NUM_DOMAINS - 1,
                                env->tcb);
+    /* Reset our domain */
+    seL4_DomainSet_Set(env->domain, 0, env->tcb);
```
kent-mcleod, line 152: This looks weird? Are both of these lines still needed?

Indanz: Not sure, can't remember whether it was just undoing any changes the test does, or for a specific reason.

I also don't know whether sel4test creates a new task for each tests or not. If it does, then it's redundant for sure.

kent-mcleod: I mean that the added `seL4_DomainSet_Set()` function call comes directly after another `seL4_DomainSet_Set()` function call to the same tcb. So the previous call looks like it should be removed?

I think this line should either be removed, or a comment added explaining why the the cleanup needs to set the domain first to `CONFIG_NUM_DOMAINS - 1` and then again to `0` for the same thread.

lsf37: Is this not an actual test? It is setting the domain of the current thread and then resetting it. The reset might not be needed if the thread is suspended anyway afterwards, but it does look cleaner to me to leave it in the domain it started in.

Indanz: Yes, the first line is the actual test, hence we store the result in `error`.
The second line is the cleanup.

kent-mcleod: ohhh. The way that GitHub ux was presenting the change it looked like this was a test cleanup function...

PR #156, opened 2026-06-02
Disable MCS scheduler+interrupt tests under SIM

midnightveil:
These tests fail for me in various ways when running under QEMU simulator. They fail ocassionally but also when interfered with using `stress -n 64` on an 8-core machine they tend to regularly fail; our GitHub runners are virtual machines and are often able to fail similarly.

QEMU is not cycle-accurate, so we expect some deviation with timing-related tests here. However, we should investigate more thoroughly, this is a basic fix so that we can build MCS in our basic pre-merge simulation tests on seL4 PRs.

Related: https://github.com/seL4/ci-actions/pull/233

### DIFF
--- a/apps/sel4test-tests/src/tests/interrupt.c
+++ b/apps/sel4test-tests/src/tests/interrupt.c
@@ -120,7 +120,9 @@ static int test_interrupt_notification_and_tcb_sc(env_t env)
 }
 DEFINE_TEST(INTERRUPT0003,
             "Test interrupts with scheduling context donation from notification object and without (two clients)",
-            test_interrupt_notification_and_tcb_sc, config_set(CONFIG_HAVE_TIMER) &&config_set(CONFIG_KERNEL_MCS));
+            test_interrupt_notification_and_tcb_sc, config_set(CONFIG_HAVE_TIMER) &&config_set(CONFIG_KERNEL_MCS)
+            /* This test is flakey under QEMU simulation: see https://github.com/seL4/ci-actions/pull/233 */
+            && !config_set(CONFIG_SIMULATION));
 
 /* test that if niether the thread or notification object have a scheduling context, nothing happens */
 static int test_interrupt_no_sc(env_t env)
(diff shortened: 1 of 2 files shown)


lsf37 (approved):
Interesting that the IRQ ones are flaky, I'd have expected them to work. But let's disable everything that is unstable first and then look more deeply afterwards.

lsf37:
Can you try if moving the comment to after/before the whole set makes the style checker not go crazy? If not, we'll have to use the bad style, otherwise it'll just always complain about this file.

midnightveil:
Oh, weird. I ran the style checker locally and it asked for no changes.... I guess I must have it set up wrong somehow. Will fix...

EDIT: Oh, the style checker is complaining about the `&&` positioning. I guess that's why everywhere is using `&&config` instead of an extra space. Nothing to do with the comment.

lsf37:
> Oh, weird. I ran the style checker locally and it asked for no changes.... I guess I must have it set up wrong somehow. Will fix...
> 
> EDIT: Oh, the style checker is complaining about the `&&` positioning. I guess that's why everywhere is using `&&config` instead of an extra space. Nothing to do with the comment.

Yes, I was just hoping that it did that because of the comment, but it appears that it just always gets that one wrong.

apps/sel4test-tests/src/tests/interrupt.c
```diff
@@ -120,7 +120,9 @@ static int test_interrupt_notification_and_tcb_sc(env_t env)
 }
 DEFINE_TEST(INTERRUPT0003,
             "Test interrupts with scheduling context donation from notification object and without (two clients)",
-            test_interrupt_notification_and_tcb_sc, config_set(CONFIG_HAVE_TIMER) &&config_set(CONFIG_KERNEL_MCS));
+            test_interrupt_notification_and_tcb_sc, config_set(CONFIG_HAVE_TIMER) &&config_set(CONFIG_KERNEL_MCS)
+            /* This test is flakey under QEMU simulation: see https://github.com/seL4/ci-actions/pull/233 */
+            && !config_set(CONFIG_SIMULATION));
```
Indanz, line 125: This one might work if we first do the `api_sc_bind` and then start the timer:

In the current version there is a race between starting the timer and binding the SC to the NF: If the timer interrupt triggers before the bind call, no SC donation will happen and the task will never wake up.

PR #157, opened 2026-06-03
Include gen_config for CONFIG_SIMULATION

midnightveil:
This wasn't included, so it wasn't defined, so it was always false and so this test always ran.

(Added in https://github.com/seL4/sel4test/pull/146 without the header)

### DIFF
--- a/apps/sel4test-tests/src/tests/faults.c
+++ b/apps/sel4test-tests/src/tests/faults.c
@@ -5,6 +5,8 @@
  */
 
 #include <autoconf.h>
+#include <sel4test-driver/gen_config.h>
+
 #include <assert.h>
 #include <stdio.h>
 #include <stdlib.h>


midnightveil:
Currently the style checker is failing with:

```diff
diff --git a/apps/sel4test-tests/src/tests/faults.c b/apps/sel4test-tests/src/tests/faults.c
index d0320f0..2d74a42 100644
--- a/apps/sel4test-tests/src/tests/faults.c
+++ b/apps/sel4test-tests/src/tests/faults.c
@@ -1063,7 +1063,7 @@ static int test_timeout_fault_nested_servers(env_t env)
     return sel4test_get_result();
 }
 DEFINE_TEST(TIMEOUTFAULT0003, "Nested timeout fault", test_timeout_fault_nested_servers,
-            config_set(CONFIG_KERNEL_MCS) && !config_set(CONFIG_SIMULATION))
+            config_set(CONFIG_KERNEL_MCS)&& !config_set(CONFIG_SIMULATION))
 
 static void vm_enter(void)
 {

```

When I fix this it then complains again, wanting it back the other way:

```diff
diff --git a/apps/sel4test-tests/src/tests/faults.c b/apps/sel4test-tests/src/tests/faults.c
index 2d74a42..d0320f0 100644
--- a/apps/sel4test-tests/src/tests/faults.c
+++ b/apps/sel4test-tests/src/tests/faults.c
@@ -1063,7 +1063,7 @@ static int test_timeout_fault_nested_servers(env_t env)
     return sel4test_get_result();
 }
 DEFINE_TEST(TIMEOUTFAULT0003, "Nested timeout fault", test_timeout_fault_nested_servers,
-            config_set(CONFIG_KERNEL_MCS)&& !config_set(CONFIG_SIMULATION))
+            config_set(CONFIG_KERNEL_MCS) && !config_set(CONFIG_SIMULATION))
 
 static void vm_enter(void)
 {
```

Thanks, astyle.

lsf37 (approved):

lsf37:
Super annoying. It's possible that it's failing to parse this properly, because it's inside a macro and there is no real actual syntactic C context that it is parsing the expression in (it's basically random text between function definitions). Not sure that helps us anything, though.

If all else fails, we can just merge it with failing style test. It will keep complaining in the future, of course.

midnightveil:
> If all else fails, we can just merge it with failing style test. It will keep complaining in the future, of course.

Yes, unfortunately. Though this seems to have been already the case pre-this PR...

PR #158, opened 2026-08-28
cache: add test for flushes outside kernel window

midnightveil:
This catches the bug where we flush addresses greater than 2^39, which is the maximum mapped vaddr in the kernel normal window, i.e. we've picked a bit of device memory outside.

### DIFF
--- a/apps/sel4test-tests/src/tests/cache.c
+++ b/apps/sel4test-tests/src/tests/cache.c
@@ -325,3 +325,111 @@ static int test_page_uncached_after_retype(env_t env)
 DEFINE_TEST(CACHEFLUSH0004, "Test that mapping a frame uncached doesn't see stale data after retype",
             test_page_uncached_after_retype,
             config_set(CONFIG_HAVE_CACHE))
+
+#ifdef CONFIG_ARCH_ARM
+static int test_flush_outside_window(env_t env)
+{
+    seL4_CPtr frame;
+    vka_object_t frame_object;
+    uintptr_t vstart;
+    vka_t *vka;
+    int error;
+
+    vka = &env->vka;
+
+    void *vaddr;
+
+    reservation_t reservation;
+
+    reservation = vspace_reserve_range(&env->vspace,
+                                       PAGE_SIZE_4K, seL4_AllRights, /* uncached */ 0, &vaddr);
+    assert(reservation.res);
+
+    vstart = (uintptr_t)vaddr;
+    assert(IS_ALIGNED(vstart, seL4_PageBits));
+
+    /* Assume that the kernel window doesn't cover the top of the device frames.
+       This is true for AArch64.
+     */
+    uintptr_t paddr = CONFIG_PADDR_USER_DEVICE_TOP - PAGE_SIZE_4K;
+    assert(IS_ALIGNED(paddr, seL4_PageBits));
+
+    /* Create a frame */
+    error = vka_alloc_frame_at(vka, PAGE_BITS_4K, paddr, &frame_object);
+    assert(!error);
+    frame = frame_object.cptr;
+    test_assert(frame != seL4_CapNull);
+
+    /* map in the frame */
+    error = vspace_map_pages_at_vaddr(&env->vspace, &frame, NULL, vaddr, 1, seL4_PageBits, reservation);
+    test_error_eq(error, seL4_NoError);
+
+    /* test frame cap based operations */
+    error = seL4_ARM_Page_Clean_Data(frame, 0, PAGE_SIZE_4K);
+#if defined(CONFIG_ARM_HYPERVISOR_SUPPORT)
+    test_eq(error, seL4_IllegalOperation);
+#else
+    test_eq(error, seL4_NoError);
+#endif
+
+    error = seL4_ARM_Page_Invalidate_Data(frame, 0, PAGE_SIZE_4K);
+#if defined(CONFIG_ARM_HYPERVISOR_SUPPORT)
+    test_eq(error, seL4_IllegalOperation);
+#else
+    test_eq(error, seL4_NoError);
+#endif
+
+    error = seL4_ARM_Page_CleanInvalidate_Data(frame, 0, PAGE_SIZE_4K);
+#if defined(CONFIG_ARM_HYPERVISOR_SUPPORT)
+    test_eq(error, seL4_IllegalOperation);
+#else
+    test_eq(error, seL4_NoError);
+#endif
+
+    error = seL4_ARM_Page_Unify_Instruction(frame, 0, PAGE_SIZE_4K);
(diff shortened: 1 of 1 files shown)


Indanz:
Hm, how come the kernel doesn't fault?

midnightveil:
Not sure... I didn't check.

midnightveil:
It could be this:

https://github.com/seL4/seL4/blob/28b8f4c40d4a48a206bedc7875c6695f6106f34b/src/arch/arm/64/kernel/vspace.c#L1481

(hopefully not though. I disagree that should be the behaviour tbh).

Indanz:
> It could be this:
> 
> https://github.com/seL4/seL4/blob/28b8f4c40d4a48a206bedc7875c6695f6106f34b/src/arch/arm/64/kernel/vspace.c#L1481

That would mean the memory you're trying to flush isn't mapped in your address space. But it is, so it can't be this.

midnightveil:
Yeah, it shouldn't be triggering. But I suppose it *could* be.

paddr_to_pptr I don't understand why it doesn't fault tbh. Even assuming 40 bit PA (so 2^40) then the kernel window is like 2^40-2^39 to 2^40-2^30 and then kernel mapping from there up a bit, with the top bit being the kernel device mappings from 2^40-2^21 up; there should only be 2 or 3 kernel devices so 2^40-2^12 should be unmapped. 44 or 48 bit PA should lie well outside that range.

midnightveil:
OK, looks like some of the runs have had kernel aborts though, looking at their logs.

ZYNQMP_debug_SMP_hyp_gcc_64 in HW Run (ZYNQMP, armv8a, gcc, 2) has.

apps/sel4test-tests/src/tests/cache.c
```diff
@@ -325,3 +325,81 @@ static int test_page_uncached_after_retype(env_t env)
...
+
+    error = seL4_ARM_Page_Unify_Instruction(frame, 0, PAGE_SIZE_4K);
+    test_eq(error, seL4_IllegalOperation);
+
+    /* test vspace based operations.
+     * on AArch64-EL1, we switch to the address space to perform CMO, so
+     * this is actually fine. for AArch64-EL2, we can't use the user address space
+     * as EL2 does use user page tables.
+     */
+#if defined(CONFIG_ARCH_AARCH64) && defined(CONFIG_ARM_HYPERVISOR_SUPPORT)
```
midnightveil: Looks like I need this check for the earlier tests too, in that for frame-based onced the EL1 is happy to use the user page tables as well. This accounts for many test failures that are not crashes, as every non-hyp aarch64 config fails to complain (since it's perfectly fine there).

Forgot about ARM32 too, it looks like it has different behaviour hyp vs non-hyp (and also it can kernel data abort as wel.

lsf37: Just to confirm: yes, the EL1 kernel uses the user page table and just checks that a page is mapped at the target address (and that it is `write` for `ARMVSpaceInvalidate_Data`).

The AArch32 version should be the same general mechanism, using the user page table for EL1 and its own for EL2. On AArch32 the kernel does not check for `write` for invalidate. There's no comment, so I'm not sure if that was missed when we added it to AArch64 or intentional.

The flush-outside-kernel window problem should otherwise be the same on AArch32 and AArch64, and it looks like that is happening.

Instead of excluding them form the test, we could explicitly check that the operation is fine for the EL1 kernels, then we get some coverage in both case.

Indanz: > On AArch32 the kernel does not check for write for invalidate.

That's probably a bug, we had something similar for AArch64 that got fixed.

lsf37: I think it might be correct for AArch32, we should check the issue discussion. I vaguely remember this coming up.

midnightveil: Architecturally we need that check for EL1:

However, from the seL4 permissions side shouldn't we check that too regardless of EL1/EL2? If one has read only access to a frame or vspace, should that imply ability to invalidate when in EL2? Should it even imply cache clean ability? (I think it should)

midnightveil:
OK so the reason why some boards are fine (e.g. IMX8MM) is that we don't test aarch64_hyp for them:

https://github.com/seL4/ci-actions/blob/5829b767d17f8918ea893e3d4f3d900ddbcb9331/seL4-platforms/platforms.yml#L214-L221

That's probably concerning... I wonder how many of those have been missed.

I'll see if there's any boards which fail now that I've actually made it not fail when it should be fine.

midnightveil:
Yep, looks like that all the failures are due to KERNEL DATA ABORT now (not tests failing) and all the full-run successes are because the hyp isn't run on that board.

(QEMU when run in simulation as hyp would fail the test, but that's because QEMU doesn't seem to emulate the side-effects of `dc cvac`; but besides we don't run QEMU hyp; EDIT: also seems like QEMU reports that there is no L3 cache and so is architecturally permitted to treat a `DC CVAU` (etc) as `NOP`, which it does)

Indanz:
> That's probably concerning... I wonder how many of those have been missed.

Probably most of the 64-bit boards that have `arch: arm`, but no `arch_hyp`: ROCKPRO64,  QUARTZ64, ARMVIRT64, IMX8MM_EVK, IMX8MP_EVK, IMX8MP_IOTGATE, TQMA8XQP1GB, ODROID_C2, RPI4, RPI4 and RPI5.

apps/sel4test-tests/src/tests/cache.c
Indanz: There are two sides to this: Cap rights and actual page mapping attributes.

For cap rights we can do what we think is sensible (e.g. I would require write for any kind of flushes), but for the page mapping attribute checks you have to follow actual hardware requirements. And it looks like AArch64 doesn't require write access, except for DC IVAC, hence the extra check. But it seems AArch32 doesn't require write for invalidate, hence the difference in code.

["G4.4.7.9 Boundary conditions for cache maintenance instructions"](https://support.arm.com/documentation/ddi0487/maa/-Part-G-The-AArch32-System-Level-Architecture/-Chapter-G4-The-AArch32-System-Level-Memory-Model/-G4-4-AArch32-cache-and-branch-predictor-support/-G4-4-7-AArch32-cache-and-branch-predictor-maintenance-instructions?lang=en#beigdicg) says:

"For VA-based cache maintenance instructions, the instructions operate on the caches regardless of the memory type and Cacheability attributes marked for the memory address in the VMSA translation table entries. This means that the effects of the cache maintenance instructions can apply regardless of:"
[..]
"The access permissions for the address accessed, other than the effect of the stage two write permission on data or unified cache invalidation instructions."

midnightveil: Yes, that's why I distinguished architecturally vs "sel4 permissions" in my previous message.

PR #159, opened 2026-09-14
stm32mp2: disable IWDG1 watchdog

bruelc:
STM32MP2: Use SMC to disable the IWDG1 watchdog, preventing unwanted watchdog reset while the test suite is running.

This fixes a long-standing annoyance that required rebuilding a new TF-A boot chain to disable the IWDG1 watchdog 

When the watchdog is started by TF-A and the corresponding OTP bit is not programmed, IWDG1 remains under secure control and can be managed through the OP-TEE watchdog SMC service.

One option would be to keep the watchdog enabled and periodically send the IWDG "PET" command through SMC. However, this would require adding watchdog refresh calls to the sel4test_run_tests() execution loop. I also evaluated using a periodic timer to perform the refres\
h automatically, but this is not practical because the timer subsystem is itself reset and tested in sel4tests.

Simply stopping IWDG1 in plat_init()  is sufficient to avoid unexpected resets


apps/sel4test-driver/src/plat/stm32mp2/plat.c
```diff
@@ -0,0 +1,43 @@
...
+
+    vka_cspace_make_path(&env->vka, smc_cap, &src_path);
+    error = vka_cnode_mint(&dest_path, &src_path, seL4_AllRights, IWDT_SMC);
+    ZF_LOGF_IF(error, "Failed to mint SMC error %d", error);
+
+    smc_args.x0 = IWDT_SMC;
+    smc_args.x1 = IWDT_ENABLE;
+    smc_args.x2 = 0; /* disable */
+
+    error = seL4_ARM_SMC_Call(dest_path.capPtr, &smc_args, &smc_results);
```
Indanz: A zero badged SMC cap gives the right to do any call, no need to mint it. You only need to do that if you want to delegate the rights to do a particular SMC calls to another task.
```suggestion
    seL4_ARM_SMCContext smc_args = {0};
    int error;

    smc_args.x0 = IWDT_SMC;
    smc_args.x1 = IWDT_ENABLE;
    smc_args.x2 = 0; /* disable */

    error = seL4_ARM_SMC_Call(seL4_CapSMC, &smc_args, &smc_results);
```
(If using `seL4_CapSMC` directly doesn't work you may need to retain the `smc_cap` code.)

bruelc: great , I used the mint from the other smc tests in the suite, but indeed that works without it and using seL4_CapSMC directly. that makes the SMC_Call even more trivial. thank you

Indanz (approved):
Thanks, this should make it possible to enable the stm boards in CI now.

Please add a comment that we are disabling the watchdog, so it's easier to see what's going on for people reading this.

Do the `ZF_LOGF_IF` work? I mean, do we have functional printing at this point? If not, better to remove it.

bruelc:
> Thanks, this should make it possible to enable the stm boards in CI now.
> 
> Please add a comment that we are disabling the watchdog, so it's easier to see what's going on for people reading this.
> 
sure

> Do the `ZF_LOGF_IF` work? I mean, do we have functional printing at this point? If not, better to remove it.

printf works at that point. But on a second though we can ignore the result if WDT is not configuredin op-tee or already disabled.

midnightveil:
Is this the right repo for this to be in? One couldn't run sel4bench because it's in sel4test; this should be in utils_libs or seL4_libs or something else, right?

Also, since SMC calls aren't enabled by default, we'll need to probably enable SMC calls for the STM32MP platform in the seL4 repository. I'd suggest at least printing a big old warning if they're not enabled, or even failing the build, because it's kind of hard to use a platform that watchdog panics...

Indanz:
sel4bench probably needs the same call during init, but it might be fast enough to finish in time (still annoying if it reboots all the time though). Maybe there is some shared library where you can add the disable watchdog function and call it from both sel4test and sel4bench platform init, but not sure that's worth the hassle compared to just duplicating the code for sel4bench. If there is a platform init code in util_libs/libplatsupport, that might be the place.

bruelc:
> Is this the right repo for this to be in? One couldn't run sel4bench because it's in sel4test; this should be in utils_libs or seL4_libs or something else, right?

we could provide a complete lib API, not only disable, but also set-timeout, kick, ... What do you advise as a hosting place to use native seL4 calls ? maybe libsel4platsuport/src/plat/stm32mp2 ? 

> 
> Also, since SMC calls aren't enabled by default, we'll need to probably enable SMC calls for the STM32MP platform in the seL4 repository. I'd suggest at least printing a big old warning if they're not enabled, or even failing the build, because it's kind of hard to use a platform that watchdog panics...

OK, we use this platform with a linux vm doing the WD refresh, but I agree for native apps.

thank you

bruelc:
> sel4bench probably needs the same call during init, but it might be fast enough to finish in time (still annoying if it reboots all the time though). Maybe there is some shared library where you can add the disable watchdog function and call it from both sel4test and sel4bench platform init, but not sure that's worth the hassle compared to just duplicating the code for sel4bench. If there is a platform init code in util_libs/libplatsupport, that might be the place.

sel4bench runs ok, but maybe I was just lucky. 
is util_libs intended to be used with native seL4 ?  . I see no include <sel4/sel4.h>, just simple

bruelc:
Tested the following API
:
libsel4platsupport/plat_include/stm32mp2/iwdg.h:  extern seL4_Error stm32_iwdg_enable(bool enable);

libsel4platsupport/src/plat/stm32mp2/iwdg.c :
#include <sel4/sel4.h>
seL4_Error stm32_iwdg_enable(bool enable)
{
#ifdef CONFIG_ALLOW_SMC_CALLS
...
#else
    return 0;
#endif
}

sel4test plat.c can just include iwdg.h and call error = stm32_iwdg_enable(false);

I'll push to libsel4platsupport repo if acceptable

thank you

bruelc:
I pushed a rfc proposal in https://github.com/seL4/seL4_libs/pull/117. Comments are very welcome to improve the API namespace or organisation,   thank you,

midnightveil (approved):
Could you also enable SMC calls by default for the STM32 in the seL4 repo?

Indanz:
> Could you also enable SMC calls by default for the STM32 in the seL4 repo?

What did you have in mind here? Do you mean in src/plat/stm32mp2/config.cmake? That is discouraged, because that's user policy and would make it impossible for the user to override the setting.

Agreed that it should be enabled in sel4test and sel4bench by default when the watchdog disable calls are added though.

midnightveil:
> What did you have in mind here? Do you mean in src/plat/stm32mp2/config.cmake? That is discouraged, because that's user policy and would make it impossible for the user to override the setting.

True. Maybe it's different fine for just sel4test and sel4bench... Basically I'm saying that we need to enable it to make our tests work some way or another.

Alternatively we could make SMC calls default to on now that they are part of the verified config, but that's probably a bigger change...

Indanz:
Oops, thought this was the seL4_Libs PR when merging. Now we need a separate PR to enable `KernelAllowSMCCalls` for stm32.

bruelc:
Wouldn’t this break the verification ?

midnightveil:
> Wouldn’t this break the verification ?

SMC calls are part of the verified kernel configuration and are verified (but I don't think we say anything about what happens when a SMC call happens - @lsf37 can comment more on what is promised), but the verified config supports having it enabled.

Indanz:
The SMC cap and syscall code is verified. But once you do a SMC call, anything can happen, so that can break the proof assumptions. This is no different in that regard than DMA without SMMU/IOMMU: Device untyped memory code is proven, but using it to setup a DMA device can break assumptions made in the proofs.

Edit: Some more clarification: The proofs treat an SMC call as an opaque machine op, they don't model what actually happens.

Indanz:
Merging this broke the build with:

```
 /github/workspace/projects/sel4test/apps/sel4test-driver/src/plat/stm32mp2/plat.c:7:10: fatal error: 'iwdg.h' file not found
      7 | #include <iwdg.h>
```
Edit: It compiles for me locally with gcc, maybe it's just clang that's broken for some reason.

lsf37:
> The SMC cap and syscall code is verified. But once you do a SMC call, anything can happen, so that can break the proof assumptions. This is no different in that regard than DMA without SMMU/IOMMU: Device untyped memory code is proven, but using it to setup a DMA device can break assumptions made in the proofs.
> 
> Edit: Some more clarification: The proofs treat an SMC call as an opaque machine op, they don't model what actually happens.

Yes, that's right. The verification allows SMC calls, but pretends that SMC calls will not change anything in the machine state that verification is interested in (e.g. page tables, TLB, memory content visible to the kernel). This is very similar to the DMA assumption.

Of course it is easily possible to break everything when using SMC calls, so they should be used with care. Switching off/on a watchdog timer is exactly what it is designed for, though, so this use is fine.

bruelc:
> Merging this broke the build with:
> 
> ```
>  /github/workspace/projects/sel4test/apps/sel4test-driver/src/plat/stm32mp2/plat.c:7:10: fatal error: 'iwdg.h' file not found
>       7 | #include <iwdg.h>
> ```
> 
> Edit: It compiles for me locally with gcc, maybe it's just clang that's broken for some reason.

I haven't tried with clang. I'll try to check let me know if you find something first

Indanz:
I think it was because I merged them in the wrong order like you said, it should be resolved now.

bruelc:
> > What did you have in mind here? Do you mean in src/plat/stm32mp2/config.cmake? That is discouraged, because that's user policy and would make it impossible for the user to override the setting.
> 
> True. Maybe it's different fine for just sel4test and sel4bench... Basically I'm saying that we need to enable it to make our tests work some way or another.
> 
we could just have in config.cmake

 if(NOT DEFINED KernelAllowSMCCalls)
    set(KernelAllowSMCCalls ON)
  endif()

override-able with -DKernelAllowSMCCalls=FALSE

Indanz:
If you mean the kernel config.cmake, then we can as well change the default from off to on.

bruelc:
> If you mean the kernel config.cmake, then we can as well change the default from off to on.

just the platform cmake. src/plat/stm32mp2/config.cmake.

I'm not confident about changing the global config_option for other arm targets :) I prefer to let this to the global maintainer

Indanz:
Ah, Julia proposed that before I think, and I replied that doing that is discouraged, because that's user policy which shouldn't be in the platform config file.

Changing the default from off to on globally might cause unexpected behaviour for existing projects. No, I think it has to be a project setting in sel4test and sel4bench, I don't see any other solution without too big downsides.

bruelc:
> Ah, Julia proposed that before I think, and I replied that doing that is discouraged, because that's user policy which shouldn't be in the platform config file.
> 
> Changing the default from off to on globally might cause unexpected behaviour for existing projects. No, I think it has to be a project setting in sel4test and sel4bench, I don't see any other solution without too big downsides.

Sorry, I understood you to mean allowing the setting to be overridden from CMake (-DKernelAllowSMCCall=FALSE). Setting it per project is indeed another story. Agree

midnightveil:
@bruelc Do you have time to look into boot failures with MCS and SMP?

https://github.com/seL4/seL4/actions/runs/35223107957/job/105212685497

It looks like to me by the assertions that are going off that CVNTOFF is not being reset to 0 for each core, as the numbers reporting in the clock sync test are very high. I'm unsure why that would be, however, as it works on all of our other platforms.

bruelc:
> @bruelc Do you have time to look into boot failures with MCS and SMP?
> 
> https://github.com/seL4/seL4/actions/runs/35223107957/job/105212685497
> 
I hit a different assertion that the one in the log

clock_sync_test[1]: t0 = 4588754202162161936, t = 17598615406297974290, td = 13009861204135812354
seL4 failed assertion 't0 <= margin + t && t <= t0 + margin' at /local/build/Sel4/sel4_test_master/kernel/src/kernel/boot.c:602 in function clock_sync_test
halting...

> It looks like to me by the assertions that are going off that CVNTOFF is not being reset to 0 for each core, as the numbers reporting in the clock sync test are very high. I'm unsure why that would be, however, as it works on all of our other platforms.

yes, possible I'm not sure that the  cntvct_el0 reset value is defined in the architecture. I'll have a look

midnightveil:
> yes, possible I'm not sure that the cntvct_el0 reset value is defined in the architecture. I'll have a look

Hm, maybe. We do set CVNTOFF to zero to make everything zero for them all, but maybe that doesn't do the right thing?

bruelc:
but does it matter ? I don't know the generic timer architecture, but this register should be derived from a counter shared by all cores. maybe this is the offset register that needs to be reset

bruelc:
works fine in EL2... CNTVCT_EL0 is perfectly synchronized between cores.  my only guess is that CNTVOFF_EL2 is not correctly initialized when booting in EL1 but that should be done by TF-A. 

edit: wrong hypothesis. since there is no timer delta in EL2, CNTVOFF_EL2 is OK

Indanz:
> works fine in EL2... CNTVCT_EL0 is perfectly synchronized between cores. my only guess is that CNTVOFF_EL2 is not correctly initialized when booting in EL1 but that should be done by TF-A.

We reset that to zero in Elfloader [here](https://github.com/seL4/seL4_tools/blob/de2de2fc6348889161966a72f65ccdf5553da1bf/elfloader-tool/include/arch-arm/64/mode/arm_generic_timer.h#L23).

However, maybe that's broken for EFI boot or something? Can you add a printf there to see whether that function is actually called?

If it isn't, then maybe something goes wrong with loading drivers from DTB in Elfloader and the `ELFLOADER_DRIVER` construction doesn't work for some reason. Maybe the `_driver_list` section gets garbage collected?

Indanz:
Okay, I discovered that the timer is missing from the generated file `elfloader/gen_headers/devices_gen.h`, for a local build I get:

```c
#ifdef DRIVER_COMMON
struct elfloader_device elfloader_devices[] = {
    {
        /* /psci */
        .compat = "arm,psci-1.0",
        .region_bases = {
            (void *)0,
        },
    },
    {
        /* /soc@0/bus@42080000/serial@400e0000 */
        .compat = "st,stm32h7-uart",
        .region_bases = {
            (void *)0x400e0000,
        },
    },
};
```
So whatever generates this is missing the timer for some reason.

Indanz:
Ah, `&{/timer}` is missing from `seL4,elfloader-devices` in src/plat/stm32mp2/overlay-stm32mp2.dts.

Indanz:
PR: https://github.com/seL4/seL4/pull/1753

PR #160, opened 2026-09-17
stm32mp2: enable SMC calls by default for wdt

midnightveil:
This is necessary as otherwise we cannot actually control or kick the watchdog timer.


Indanz (approved):
I double checked that the kernel platform name is indeed just MP2.

